PR #1, opened 2022-01-30
Run Unit Tests

tpreusse:
The backend module test folder somehow had to be renamed to testing for turbo to recognise it.

`@orbiting/backend-modules-test` is now responsible to run all the backend unit tests. The jest config was moved there adapted to the new path.

Also the react version had to be bumped in the backend for styleguide test to run properly. Publikator build has been disabled for now until memo and styleguide are compatible.


patrickvenetz (approved):
LGTM

PR #13, opened 2022-02-02
fix: Active Debate section - DisucssionIcon

trm217:
# Description

Remove the discussion-icon next to the "Aktive Debatten" heading. It is removed for two reasons:
1. Consistent headings on the page
2. The Icon had faulty styling when using darkmode (see screenshot below)
image

Instead of removing the icon completely, it might an option to add it next to the green numbers for each active debate. That way it would be consistent compared to how the amount of comments are displayed, for example, on the front. (See screenshot below for a comparison) 
image

# Screenshot

## Before

image

## After

image

### DIFF
--- a/apps/www/pages/dialog.js
+++ b/apps/www/pages/dialog.js
@@ -7,7 +7,8 @@ import {
   Editorial,
   inQuotes,
   Interaction,
-  mediaQueries
+  mediaQueries,
+  useColorContext
 } from '@project-r/styleguide'
 import {
   CDN_FRONTEND_BASE_URL,
@@ -75,6 +76,7 @@ const SUPPORTED_TABS = ['general', 'article']
 const DialogContent = ({ tab, activeDiscussionId, serverContext }) => {
   const { t } = useTranslation()
   const { query } = useRouter()
+  const [colorScheme] = useColorContext()
 
   const discussionContext = useDiscussion()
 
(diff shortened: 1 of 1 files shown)


trm217:
@tpreusse @ovbm What are your thoughts on this? Should we perhaps show the icon next to the comment-count?

ovbm:
i think it's too busy. Prefer the current solution.

ovbm (approved):

tpreusse:
The correct solution would be to just fix the regression I've introduced when cleaning up the discussion refactor:
https://github.com/republik/plattform/commit/8ad0dd8c82d2ca81e76eabecc888a1bfe8d13d75#diff-1bc5e89a7c49b7baf8860e167e1b7c9fea1c79508ac6856e7336f4901ce4de81L188

Either go back to the old deprecated style that worked:
```js
<DiscussionIcon size={24} fill={colors.primary} />
```

Or go to the new style:
```js
<DiscussionIcon size={24} {...colorScheme.set('fill', 'primary')} />
```

Or introduce a `fillColorName` prop on `DiscussionIcon` like on `IconButton` and then do something like this:
```js
<DiscussionIcon size={24} fillColorName='primary' />
```

trm217:
@tpreusse From a UI view the way it was seems weird. From a consistency point of view, wouldn't it make sense to make sure it's used in the same way as on the front? (We also use the icon + comment count in other places)

tpreusse:
@trm217 like I said before we should not change design at random. The dialog page was designed as a whole a few years ago and the icon makes sense the way it is used. Repeating things in a list or table is a bad pattern. This was actually even discussed back then. If you want to redesign the page: yes but please separat from a bug fix and with clear specs & goals.

tpreusse (approved):

PR #16, opened 2022-02-02
chore(commitlint): ignore node-version requirements when running commitlint

trm217:
@patrickvenetz could you please check if this also fixes the issues that you experienced?


patrickvenetz (approved):
LGTM

PR #18, opened 2022-02-03
fix(Portrait): darkmode compatibility for drop area

ovbm:
https://github.com/orbiting/republik-frontend/issues/679

before
Screenshot 2022-02-03 at 11 00 50

after
Screenshot 2022-02-03 at 11 00 32

### DIFF
--- a/apps/www/components/Profile/Portrait.js
+++ b/apps/www/components/Profile/Portrait.js
@@ -4,24 +4,23 @@ import Dropzone from 'react-dropzone'
 
 import withT from '../../lib/withT'
 
-import { CloseIcon } from '@project-r/styleguide'
+import { CloseIcon, useColorContext } from '@project-r/styleguide'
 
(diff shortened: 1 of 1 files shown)


apps/www/components/Profile/Portrait.js
```diff
@@ -77,27 +76,30 @@ const PortraitSelector = withT(
...
 
+    const [colorScheme] = useColorContext()
+
     const preview = isEditing && values.portraitPreview
     const imgUrl =
       values.portrait !== undefined ? values.portraitPreview : user.portrait
     const img = (
       <span
         {...styles.img}
+        {...colorScheme.set('backgroundColor', 'hover')}
```
ovbm, line 99: Most of the PR ist just linter. This is the relevant change, including removal of background-color on line 12

patrickvenetz (approved):
LGTM

PR #19, opened 2022-02-03
fix(Nav): only show search field if has activeMembership

### LINKED ISSUE #8: Search field visibel to logged in non-members
As a user without an active membership, opening navigation in top right corner shows search input field.

I didn't expect an input field visible that I can not use.

Bildschirmfoto 2022-01-26 um 14 36 45

ovbm:
closes #8 

@patrickvenetz - couldn't test because my backend is whack. Could you test to see if behavior is as expected now?

### DIFF
--- a/apps/www/components/Frame/Popover/Nav.js
+++ b/apps/www/components/Frame/Popover/Nav.js
@@ -31,13 +31,12 @@ const Nav = ({
   inIOS,
   inNativeIOSApp,
   isMember,
-  hasActiveMembership,
+  hasAccess,
   onSearchSubmit,
 }) => {
   const [colorScheme] = useColorContext()
   const active = router.asPath
   const hasExpandedRef = useRef(expanded)
-  const hasAccessGrants = me?.accessGrants && !!me?.accessGrants.length
   if (expanded) {
     hasExpandedRef.current = true
   }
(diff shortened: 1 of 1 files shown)


patrickvenetz (changes requested):
User with an active access grant should see search field, too.

apps/www/components/Frame/Popover/Nav.js
```diff
@@ -58,7 +60,7 @@ const Nav = ({
                 </Button>
               </Link>
             )}
-            {me && (
+            {(hasActiveMembership || hasAccessGrants) && (
```
patrickvenetz, line 63: Would use `hasAccess` instead, similar to onboarding hint check on account page.

`hasAccess` is provided by `withMe`. It checks for role `member`, which indicates whether you have access due to a membership or an access grant.

ovbm: that's so weird. because hasAccessGrant is used in various other places. Probably a case that could be cleaned a bit.

ovbm: What confused me: 
hasAccess
isMember
Are equivalent.
Some places use isMember, some isAccess. very confusing

patrickvenetz: It is confusing.

Props avail largely depend on what `with…` a component is wrapped in e.g. `isMember` is provided by `withMembership` – albeit pretty much checking the same: has a user a `member` role.

patrickvenetz (approved):
LGTM

PR #22, opened 2022-02-03
Update userCanComment resolver to return false if the discussion has been closed.

### LINKED ISSUE #5: Reply button in closed discussions clickable
I can click on reply despite discussion being closed. It then report discussion as being closed.

I expected not being able to reply.

Bildschirmfoto 2022-01-31 um 11 48 22

[Sample discussion on republik.ch](https://www.republik.ch/vote/juli19/diskutieren)

trm217:
Closes #5 

# Description

Don't pass the reply-handler to the CommentNode if the discussion is closed. (Line 112)

### DIFF
--- a/apps/www/components/Discussion/CommentContainers/CommentContainer.tsx
+++ b/apps/www/components/Discussion/CommentContainers/CommentContainer.tsx
@@ -3,12 +3,12 @@ import React, {
   useCallback,
   useEffect,
   useMemo,
-  useState
+  useState,
 } from 'react'
 import {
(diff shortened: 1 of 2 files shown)


patrickvenetz (commented):
Now that I see `discussion.userCanComment` is the culprit, I tend to argue to change api instead of www. A user can't comment on closed discussions, thus `Discussion.userCanComment` should return `false` at once.

trm217:
Feel free to revert my commit and update the api.

PR #25, opened 2022-02-04
fix(Discussion): unsafe localStorage.getItem

tpreusse:
Some versions of safari have a `localStorage` object but no `getItem` function.

While some `readDraft` were wrapped in a try block, a `readDiscussionCommentDraft` was not (my fault). However I think in code paths we ignore broken local storage we should do the try catch everywhere we interact with it instead of assuming that our function is wrapped in a silent try catch as a whole. Wrapping the whole function also hides coding errors that are unrelated to local storage being available.  

Screenshot 2022-02-04 at 12 32 52
Screenshot 2022-02-04 at 12 33 11

### DIFF
--- a/packages/styleguide/src/components/Discussion/Composer/CommentComposer.js
+++ b/packages/styleguide/src/components/Discussion/Composer/CommentComposer.js
@@ -154,11 +154,7 @@ export const CommentComposer = ({
     if (initialText) {
       return initialText
     } else if (!isEditing) {
-      try {
-        return readDraft(discussionId, parentId) ?? ''
-      } catch (e) {
-        return ''
-      }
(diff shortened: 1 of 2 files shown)


trm217 (approved):
LGTM
I am unsure however if we really shouldn't emit any logs when an error is caught. 🤔

tpreusse:
@trm217

I theory warning on write fail to the user would be good: no local storage support – no draft saving. However two caveats:

1. we don't actively communicate that we are saving drafts
2. all modern browser decided to implement local storage in private mode but erase it after closing the window. this happened because nytimes and others startet to detect private mode via local storage.

At least the reads can definitely be silent fails. In other places were saving is vital – e.g. publikator – we do warn if it fails (afaik).

PR #27, opened 2022-02-04
Create new packages to share ESLint configurations across all apps & packages.

trm217:
# Description

Create the following two new packages:
- eslint-config-frontend
- eslint-config-backend

The ESlint-dependency was moved into the root of the repository.
On the root-level the eslintrc defines that prettier should be used and acts as the baseline for all other eslint-configurations.


trm217:
A package was initialised in `packages/backend-modules` to provide a eslint-config for all nested backend-modules.

tpreusse (approved):

PR #28, opened 2022-02-07
build(styleguide): rm rollup, just use tsc

tpreusse:
This fixes issues where a partially built styleguide causes next server to fail and throw errors.

Verified with `yarn analyze` to not have any impact on bundle size in `apps/www`.


packages/styleguide/tsconfig.json
```diff
@@ -17,10 +17,8 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "preserve",
-    "allowSyntheticDefaultImports": true,
-    "declaration": true,
-    "declarationDir": "dist/types"
+    "jsx": "react",
```
trm217, line 20: I think the correct value to use with React 17 would be `"react-jsx"`.
https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#react-17-jsx-factories

annatraussnig (approved):
The code looks very codeful indeed, and I don't seem to get the error anymore. 🐑?

PR #31, opened 2022-02-09
Audio fixes

ovbm:
- allow to suppress `<SectionNav />` and `<SectionFeed />` on audio page with two new meta props: `hideFeed` and `hideSectionNav`
- fixes TeaserCarousel padding (see before after below)
- `feedQueryVariables` are now also used in <FormatFeed /> analogous to <SectionFeed />
- new Field in Publikator to set `kind` of audioSource

next Steps:
- [x] Backend implements filter option based on: meta.audioSource.kind
- [ ] Test on staging

---
Enables "Vorgelesen"-Page, that shows all Articles in a Feed that have an audioSource.kind === readAloud:

Screenshot 2022-02-15 at 15 02 55

---

before
Screenshot 2022-02-09 at 10 27 06

after
Screenshot 2022-02-09 at 10 26 19


packages/styleguide/src/components/TeaserCarousel/Tile.js
```diff
@@ -131,16 +135,16 @@ const Tile = ({
       borderWidth: outline && 1,
       borderStyle: outline && 'solid',
       cursor: onClick ? 'pointer' : 'default',
-      padding: bigger ? '0 0 20px 0' : '30px 15px',
-      alignItems: bigger ? 'flex-start' : 'center'
+      padding: bigger & !outline ? '0 0 20px 0' : '30px 15px',
```
ovbm, line 138: relevant change

apps/www/components/Sections/SectionFeed.js
```diff
@@ -90,8 +89,14 @@ const getFeedDocuments = gql`
...
-const SectionFeed = ({ t, formats, variablesAsString }) => {
-  if (!variablesAsString && !(formats && formats.length)) {
+const SectionFeed = ({
+  t,
+  formats,
+  variables = {
+    filter: { formats, feed: true },
+  },
+}) => {
+  if (!variables && !(formats && formats.length)) {
```
tpreusse, line 99: This check no longer works because `variables` with the default value will never be falsy. Easies way: keep it like before where the default value is constructed after this check.

packages/styleguide/src/components/TeaserCarousel/Tile.js
```diff
@@ -135,7 +135,7 @@ const Tile = ({
       borderWidth: outline && 1,
       borderStyle: outline && 'solid',
       cursor: onClick ? 'pointer' : 'default',
-      padding: bigger ? '0 0 20px 0' : '30px 15px',
+      padding: bigger & !outline ? '0 0 20px 0' : '30px 15px',
```
tpreusse, line 138: In my mind we would have done the padding only around the text and keep the images as big as possible: 
Screenshot 2022-02-10 at 13 03 05

Could you add an example in the styleguide with the proposed content by Sven and check how a while row looks? Would also be good to have this combination documented in the styleguide.

apps/www/components/Sections/SectionFeed.js
ovbm: done

tpreusse (approved):

tpreusse (approved):

PR #32, opened 2022-02-09
Styleguide emit type declarations

trm217:
## Description

Back when we used rollup, a plugin handled the generation of the type declarations.
Due to the removal of rollup, types were no longer emitted. 

With this PR the config was updated to emit type declarations again. 
Also source-maps for both the JS and the TS declaration will now be generated when running dev:lib.


trm217:
Merci Thomas! 🚀

PR #35, opened 2022-02-14
feat(documents): prepend format external base url if external base url differs

patrickvenetz:
Prepends a given format `externalBaseUrl` on all links if `externalBaseUrl` in requested document differs.


patrickvenetz:
Commits wandered into https://github.com/republik/plattform/pull/29 as we are a collective now.

PR #37, opened 2022-02-17
Fix Statement Focus

tpreusse:
This introduces a 404 status code and error message if a community focus id is not returned.

Additionally focus are no longer returned from the backend if the user no longer has an active member role. This matches the behaviour in the list where past members are already filtered out.

Screenshot 2022-02-17 at 13 36 46

### DIFF
--- a/apps/www/components/Testimonial/List.js
+++ b/apps/www/components/Testimonial/List.js
@@ -29,6 +29,7 @@ import {
   useColorContext,
 } from '@project-r/styleguide'
 import { withRouter } from 'next/router'
+import ErrorMessage from '../ErrorMessage'
 
 const { P } = Interaction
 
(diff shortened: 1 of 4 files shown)


tpreusse:
The limit to active members was introduced 3 years ago: https://github.com/republik/plattform/commit/4c1d201bd7605f9372cbb335fd93e7eab736b9ef#diff-dd1ad618a7aedad3de2847ada46375eedb4a61391693dc944e5070fc5d1c9d05

patrickvenetz:
Stupid me. (Misread `${!membershipAfter ? 'AND u.roles @> \'["member"]\'' : ''}`.)

apps/www/components/Testimonial/List.js
```diff
@@ -347,7 +349,12 @@ export class List extends Component {
         render={() => {
           const items = []
           const lastIndex = statements.length - 1
-          const focusItem = focus && statements[0]
+          const focusItem = statements[0]?.id === focus ? statements[0] : null
```
patrickvenetz, line 352:
```suggestion
          const focusItem = statements[0]?.id === focus && statements[0]
```

patrickvenetz (approved):

PR #38, opened 2022-02-18
Add suppport for Apple- / Google-Pay to the checkout

trm217:
# Tasks

## Checkout

- [x] Cleanup code
- [x] Improve logic of the first paymet-request initialization
- [x] Add a loading status to usePaymentRequest-hook and don't render submit-button while status=loading.
- [x] Normalize address data received from Payment UI
- [x] Delay complete('sucess') call until payPledege has succeeded.
- [x] Define final style of the Apple Pay button
- [x] Add Google Pay (adding button + slight adaptations to functionalities that handle Apple Pay)
- [x] Add message if payment-method is temporarily unavailable

## Konto
- [x] ~Show Apple Pay icon next to the Credit-Card Icon in~ 
- [x] ~Ensure that Apple Pay is not shown as an option when adding adding a new payment-source~ (No changes necessary, Apple Pay is only available in the PaymentForm) 

## Backend
- [x] ~Update transactional-emails to mention when a wallet-card is used.~ (Since we don't especially differentiate between Apple Pay and card payments, this isn't really needed)

## Finishing touches
- [x] delete console.debug statements in Submit.js and usePaymentRequest.tsx


apps/www/pages/angebote.js
```diff
@@ -89,4 +89,4 @@ class PledgePage extends Component {
   }
 }
 
-export default withDefaultSSR(withRouter(PledgePage))
+export default withRouter(PledgePage)
```
tpreusse, line 92: please re-add

apps/www/components/Payment/Form/ApplePay.tsx
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react'
+
+// Check if the website is running in a secure document.
+function isInSecureDocument(): boolean {
+  return (
+    typeof window !== 'undefined' && window?.location?.protocol === 'https:'
```
tpreusse, line 6:
```suggestion
    typeof window !== 'undefined' && window.location.protocol === 'https:'
```

apps/www/components/Payment/Form/ApplePay.tsx
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react'
+
+// Check if the website is running in a secure document.
+function isInSecureDocument(): boolean {
```
tpreusse, line 4: `isInSecureWindow` or `isInSecureClientWindow` would make it more obvious that this only return true on the client.

apps/www/components/Payment/Form/ApplePay.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+// Declare ApplePaySession property to prevent TypeScript errors
+declare global {
+  interface Window {
+    ApplePaySession: ApplePaySession | undefined
+  }
+}
+
+function isApplePayAvailable(): boolean {
+  if (!window?.ApplePaySession) {
```
tpreusse, line 18:
```suggestion
  if (!window.ApplePaySession) {
```

apps/www/components/Payment/PaymentRequest/PaymentRequestEventHelper.ts
```diff
@@ -0,0 +1,48 @@
+import {
+  PaymentMethod,
```
tpreusse, line 2: unused?

apps/www/components/Payment/PaymentRequest/PaymentRequestEventHelper.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function getPayerInformationFromEvent(
+  event: PaymentRequestPaymentMethodEvent,
+): PayerInformation {
+  const [firstName, lastName] = event.payerName.split(' ').map((s) => s.trim())
+
+  return {
+    email: event.payerEmail,
+    firstName,
+    lastName,
```
tpreusse, line 30: Let's save everything. And I'd gamble that multiple first names are more common than space separated last names like `von Wunschheim`:

```suggestion
  const [...firstNames, lastName] = event.payerName.split(' ')

  return {
    email: event.payerEmail,
    firstName: firstNames.join(' ').trim(),
    lastName: lastName.trim(),
```

apps/www/components/Pledge/Submit.js
```diff
@@ -129,17 +132,60 @@ const SubmitWithHooks = (props) => {
...
+    return enhancedPaymentMethods
+  }, [paymentMethods, isApplePayAvailable])
+
+  const paymentRequest = useStripePaymentRequest({
+    total: {
+      amount: props.total ?? 0,
+      label: props.query.package ?? 'NO VALUE',
+    },
+    requestShipping: props?.requireShippingAddress ?? false,
+    shippingOptions: props?.requireShippingAddress
```
tpreusse, line 155:
```suggestion
    requestShipping: props.requireShippingAddress || false,
    shippingOptions: props.requireShippingAddress
```

apps/www/components/Pledge/Submit.js
```diff
@@ -129,17 +132,60 @@ const SubmitWithHooks = (props) => {
...
+      enhancedPaymentMethods = ['STRIPE-APPLEPAY', ...enhancedPaymentMethods]
+    }
+
+    return enhancedPaymentMethods
+  }, [paymentMethods, isApplePayAvailable])
+
+  const paymentRequest = useStripePaymentRequest({
+    total: {
+      amount: props.total ?? 0,
+      label: props.query.package ?? 'NO VALUE',
```
tpreusse, line 152: let's start with the package title
```
      label: t.first([`package/${props.query.package}/title`, 'package/choose']),
```

apps/www/components/Pledge/Submit.js
```diff
@@ -511,14 +634,15 @@ class Submit extends Component {
...
+
   getAutoPayValue() {
     const { forceAutoPay, options } = this.props
     const {
       values: { paymentMethod },
       autoPay,
     } = this.state
 
-    if (paymentMethod !== 'STRIPE') {
+    if (paymentMethod !== 'STRIPE' && !paymentMethod?.startsWith('STRIPE-')) {
```
tpreusse, line 645:
```suggestion
    if (!paymentMethod || !paymentMethod.startsWith('STRIPE')) {
```

apps/www/components/Pledge/Submit.js
```diff
@@ -532,11 +656,12 @@ class Submit extends Component {
     }
     return autoPay
   }
+
   renderAutoPay() {
     const {
       values: { paymentMethod },
     } = this.state
-    if (paymentMethod !== 'STRIPE') {
+    if (paymentMethod !== 'STRIPE' && !paymentMethod?.startsWith('STRIPE-')) {
```
tpreusse, line 664:
```suggestion
    if (!paymentMethod || !paymentMethod.startsWith('STRIPE')) {
```

tpreusse (commented):
maybe we should call it `STRIPE-WALLET-APPLEPAY` and then do `STRIPE-WALLET` checks instead of `STRIPE-` which seem cryptic

is the `billing_details.address` visible client side or would that be something our backend has to grab?
Screenshot 2022-02-22 at 22 40 11

apps/www/pages/angebote.js
trm217: Alright, this will however cause Apple-Pay to not be initially available on the first-render.

apps/www/components/Payment/Form/ApplePay.tsx
trm217: Wouldn't this cause an error when a Browser has not defined ApplePaySession?

trm217:
The billing address is not available in the [PaymentResponse object](https://stripe.com/docs/js/appendix/payment_response).

tpreusse:
I just tested ngrok and works on my machine too.

We need to move the «Schon mal ausgefüllt?» upwards, because after login one might also have a saved credit card, and rename it to some thing like «Haben Sie bereits ein Konto? Melden Sie sich an.»
Screenshot 2022-03-16 at 08 52 40

When choosing pay by bill we should first ask for email and name before the full address:
Screenshot 2022-03-16 at 08 53 53

Maybe we could do that for all payment options? And even switch it around that we first ask for email and then derive first name and last name from the email to pre fill name fields? (when e.g. username of email is separated by a dot and contains at least two chars on each side)

trm217:
Regarding the prefilling of the name based on the email: I wouldn't do that. Unless we have a bad implementation of the form fields, the prefilling would be suggested by every modern browser anyways. This would also cause issues with wrong prefills: for example fname and lname might be reversed in the email and thus the prefilling would be wrong, which then could easily be overseen by the user.

trm217:
image
I updated the UI as requested, I also removed some whitespace.

tpreusse (commented):
I refactored the `selectedPaymentMethod` to be available in `SubmitWithHooks` – let's move this code to the `usePaymentRequest` hook ;)

Spacing when apple pay is selected is too small, add 10 to 30 more pixels between payment method & checkboxes:

Screenshot 2022-03-17 at 13 24 49


Currently shipping address is above payment method selection: should be below (since hidden when apple pay)

Screenshot 2022-03-17 at 13 35 06

apps/www/components/Pledge/Submit.js
```diff
@@ -610,151 +898,165 @@ class Submit extends Component {
...
+              // Only render the browser API in case we're not using a browser payment API
+              !me &&
+                this.props.selectedPaymentMethod &&
+                !this.isStripeWalletPayment() && (
+                  <div {...styles.topMargin}>
+                    {contactPreface && (
+                      <div style={{ marginBottom: 40 }}>
+                        <P>{contactPreface}</P>
+                      </div>
+                    )}
```
tpreusse, line 1022: We should move this preface above the payment question:
Screenshot 2022-03-18 at 11 08 19

apps/www/components/Pledge/Submit.js
```diff
@@ -172,6 +236,110 @@ class Submit extends Component {
...
+    if (
+      this.props.selectedPaymentMethod &&
+      !this.props.selectedPaymentMethod.startsWith('STRIPE') &&
+      this.state.stripeError
+    ) {
+      this.setState(() => ({
+        stripeError: null,
+      }))
+    }
+  }
```
tpreusse, line 341: I would have imagined that we move this whole logic to the `usePaymentRequest` hook

apps/www/components/Pledge/Submit.js
```diff
@@ -172,6 +236,110 @@ class Submit extends Component {
...
+            this.setState((prevState) => ({
+              ...prevState,
+              loading: false,
+              walletError: !isGooglePayPayment
+                ? t('account/pledges/payment/methods/unavailable')
+                : null,
+              stripeError: isGooglePayPayment
+                ? t('account/pledges/payment/methods/google-pay/unavailable')
+                : null,
+            }))
```
tpreusse, line 297: IMO this logic could be made generic, for the odd case where is happens with apple pay

apps/www/components/Pledge/Submit.js
trm217: In case we do that, we would have to tear most of the Submit-component state apart and move it up to Submit Hooks. If we were to do that, we might as well refactor the entire Submit-component and convert it into a functional-component.

tpreusse: are you sure? As far as I can see it's only `loading`, `walletError`, `stripeError` – afaik our `PaymentRequestValues` already have a loading state and could be extended with two error states?

apps/www/components/Pledge/Submit.js
trm217: Done

apps/www/components/Pledge/Submit.js
trm217: Done, now works for both Apple and Google Pay (also works if other Wallets are added)

apps/www/components/Pledge/Submit.js
trm217: Done

tpreusse (approved):
Let's ship this

apps/www/components/Payment/Form/StripeWalletHelpers.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export function useIsGooglePayAvailable(): [
+  boolean,
+  Dispatch<SetStateAction<boolean>>,
+] {
+  const [isAvailable, setIsAvailable] = useState(false)
+
+  useEffect(() => {
+    // Only check for Google Pay on the client and when a secure connection is provided
+    setIsAvailable(isInSecureWindow() && isGooglePayAvailable())
```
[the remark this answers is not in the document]
trm217: Allright let's remove it.  Might cause confusion in dev though since Google Pay might show up in http.

apps/www/components/Payment/Form/StripeWalletHelpers.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+// Declare ApplePaySession property to prevent TypeScript errors
+declare global {
+  interface Window {
+    ApplePaySession: ApplePaySession | undefined
+  }
+}
+
+function isApplePayAvailable(): boolean {
+  if (!window.ApplePaySession) {
```
tpreusse, line 18: don't forget to add it here ;-)
```suggestion
  if (!isInSecureWindow() && !window.ApplePaySession) {
```

trm217: Shouldn't be an issue since ApplePaySession isn't defined on non HTTPS. But "doppelt hält besser" xD

PR #39, opened 2022-02-18
Bump required Node Version to 16

trm217:
## Description

Bumping the node-version will increase the performance of the developer environments.
Node 16 is the first Node LTS Version that has an official ARM build for MacOS thus especially people with Apple Silicon processors will benefit from this upgrade.


trm217:
@tpreusse 
Seems the tests are indeed stuck again. I tried using `yarn test --graph` to find a possible circular dependency, but I wan't able to find one. Would you mind looking into it sometime?

tpreusse:
@trm217 seems to be a turbo issue: https://github.com/vercel/turborepo/issues/766

trm217:
Lot's of Turbo related issues at the moment :/

trm217:
the mentioned issue was closed, but after bumping to turbo@1.1.5 it seems the error still persists

PR #40, opened 2022-02-21
feat(Chart): add multi line annotations and text alignment functionality

felixmichel:
Bildschirmfoto 2022-02-21 um 16 56 55

## add multi line functionality

Now annotations are always wrapped in a tspan. This could but should not be a breaking change. I also adapted the annotation position a little bit.


packages/styleguide/src/components/Chart/LineGroup.js
```diff
@@ -297,23 +297,27 @@ const LineGroup = (props) => {
...
-          textAnchor = 'end'
-          if ((range ? x2 : x1) - annotation.labelSize < 0) {
-            textAnchor = 'start'
-          }
-        }
         let tx = x1
-        if (range) {
-          if (textAnchor === 'end') {
-            tx = x2
+        if (annotation.textAlignment === undefined) {
```
tpreusse, line 301:
```suggestion
        if (!textAlignmentDict[annotation.textAlignment]) {
```

tpreusse (approved):
lgtm, two small non critical details

PR #43, opened 2022-02-23
fix: remove margin between subsequential article carousels

tpreusse:
because it came up in Slack, maybe better?

| before | after |
|---|---|
| screencapture-republik-ch-audio-2022-02-23-11_08_59 | screencapture-localhost-3010-audio-2022-02-23-11_08_04 |
| | screencapture-localhost-3010-audio-2022-02-23-11_08_31 |


ovbm (approved):
defo better, thanks!

generally: I don't like how it's white on top and then black. It should just follow darkmode pattern. Not a fan of black background for no reason (other than it's audio). But that's for another day.

PR #49, opened 2022-02-28
fix(ActionBar): flexWrap in seriesEpisode

ovbm:
Fixes overlong ActionBar when audio and reading position are both present. Scopes the fix to only concern SeriesEpisode-ActionBar
Result:
Screenshot 2022-02-28 at 12 17 06


tpreusse (approved):
LGTM

PR #51, opened 2022-03-02
Free Weekly Newsletter!

patrickvenetz:
tba.

[Notion page](https://www.notion.so/republikmagazin/Wochen-bersicht-e6f14376ac8f4ba78e5c1e08c393d79b?p=1c84697082994e5c80c78742e7e63882)


patrickvenetz:
Querying a format with processor `formatFeedSamples` will return some recently published documents despite user not being authenticated.

Configurable via ENV variables:

```
DOCUMENTS_SAMPLES_REPO_IDS=republik/format-wochenende-newsletter
DOCUMENTS_SAMPLES=3
```

Bildschirmfoto 2022-03-03 um 10 52 20

packages/backend-modules/search/lib/options.js
```diff
@@ -0,0 +1,106 @@
...
+const processors = {
+  formatFeedSamples: (args, context) => {
+    const { filter } = args
+    const { user } = context
+
+    if (
+      DOCUMENTS_SAMPLES_REPO_IDS &&
+      !user?.roles?.includes('member') &&
+      filter?.format &&
+      DOCUMENTS_SAMPLES_REPO_IDS.split(',').includes(filter?.format)
```
tpreusse, line 14:
```suggestion
      !user?.roles?.includes('member') &&
      DOCUMENTS_SAMPLES_REPO_IDS?.split(',').includes(filter?.format)
```

but should we always use the helper `userIsInRole`? And maybe do the `DOCUMENTS_SAMPLES_REPO_IDS` split just once outside of the `formatFeedSamples` process function?

tpreusse (approved):
This parts looks good and IMO should be shipped asap. The text adaption on account etc. will need to be tweaked until the last minute tomorrow and ship last minute.

patrickvenetz:
Agree with shipping.

packages/backend-modules/search/lib/options.js
patrickvenetz: Yes, `userIsInRole` is muches better and splitting ENV var outside fn call, too. Incorporated.

tpreusse (approved):
let's ship it 🚢

PR #53, opened 2022-03-03
AudioPlayer with PlaybackRate

ovbm:
Screenshot 2022-03-24 at 10 43 08

### DIFF
--- a/packages/styleguide/src/components/AudioPlayer/Player.js
+++ b/packages/styleguide/src/components/AudioPlayer/Player.js
@@ -410,7 +410,7 @@ class AudioPlayer extends Component {
       attributes = {},
       style,
       t,
-      height,
+      height: heightProp,
       closeHandler,
       download,
       scrubberPosition,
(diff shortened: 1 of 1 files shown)


trm217 (commented):
Code looks good to me on first glance 👍🏻 

One thing I am not quite fond of is using the left-right arrow-cursor when hovering over the progress-bar. I have never seen a cursor other than the pointer being used for changing the progress in audio-/video-players, therefor I think we should stick to the norm and use `cursor: pointer` as well.

~While playing around with the overlay player in the browser I noticed a weird behaviour (see the recording linked below):
After changing the progress by clicking in the progress-bar, moving the mouse cursor caused the progress to change. Can you reproduce that behaviour?~

tpreusse:
the progress moving bug after changing position is a catalog / styleguide iframe problem. can be ignore. (we need to get ride of catalog :D)

tpreusse (approved):

PR #56, opened 2022-03-04
feat(documents): process IF hasAccess in document.content

patrickvenetz:
Unwraps zone identifier `"IF"` when `data.present` is set to `hasAccess`.

**Example**

```json
{
  "identifier": "IF",
  "data": {
    "present": "hasAccess"
  },
  "children": [
    {
      "children": [
        {
          "type": "text",
          "value": "(Mitglied.)"
        }
      ],
      "type": "paragraph"
    },
    {
      "identifier": "ELSE",
      "data": {},
      "children": [
        {
          "children": [
            {
              "type": "text",
              "value": "(Kein Mitglied)"
            }
          ],
          "type": "paragraph"
        }
      ],
      "type": "zone"
    }
  ],
  "type": "zone"
}
```

When a user role has unrestricted access, zone identifier `"IF"` `children` is unwrapped. In `children` contain a zone identifier `"ELSE"`, entire zone is omitted.

```json
{
  "children": [
    {
      "type": "text",
      "value": "(Mitglied.)"
    }
  ],
  "type": "paragraph"
}
```

If user has no access, zone identifier `"ELSE"` `children` maybe be returned.

```json
{
  "children": [
    {
      "type": "text",
      "value": "(Kein Mitglied)"
    }
  ],
  "type": "paragraph"
}
```

### DIFF
--- a/packages/backend-modules/documents/lib/process.js
+++ b/packages/backend-modules/documents/lib/process.js
@@ -167,6 +167,35 @@ const processNodeModifiersInContent = (mdast, user) => {
   })
 }
 
+const processIfHasAccess = (mdast, user) => {
+  visit(mdast, 'zone', (node, index, parent) => {
+    if (node.identifier === 'IF' && node.data.present === 'hasAccess') {
+      let children = [...node.children]
+
+      const elseIndex = node.children.findIndex(
+        (child) => child.identifier === 'ELSE',
+      )
+
+      if (elseIndex >= 0) {
+        if (userIsInRoles(user, documentsRestrictToRoles)) {
+          children = [
+            ...children.slice(0, elseIndex),
+            ...children.slice(elseIndex + 1),
+          ]
+        } else {
+          children = node.children?.[elseIndex]?.children
+        }
+      }
+
+      parent.children = [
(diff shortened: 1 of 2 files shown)


packages/backend-modules/documents/lib/process.js
```diff
@@ -167,11 +167,43 @@ const processNodeModifiersInContent = (mdast, user) => {
   })
 }
 
+const processIfHasAccess = (mdast, user) => {
```
tpreusse, line 170: can we add unit tests for this function? From reading the code I feel like if block without else and or else blocks at index 0 (falsy) might not be handled correctly

patrickvenetz: Yeah, yeah, it's a bug!

patrickvenetz: Yeah, yeah, we can tescht that.

patrickvenetz: Fixed bug. Added rudimentary test.

fuenkchen (approved):
lgtm

PR #57, opened 2022-03-04
feat(publikator): toggle logged in user in preview

patrickvenetz:
Preview a document as user without restrictions (default) or without logged in.

Bildschirmfoto 2022-03-04 um 14 04 20


patrickvenetz:
(Intentionally not representing logged in user without access.)

PR #61, opened 2022-03-10
Fix focused comment not loading if it's not inside the initial comment query result

trm217:
## Description
Fix an error that caused focus-urls to break when sharing a reply.
(After spending the better part of a day trying to pin-point the issue, we managed to figure out that is caused by due to the parent-comment being filtered out since they were not assigned a topLevel value, that caused them to remain in the list of initially returned comments)


patrickvenetz (approved):
LGTM

patrickvenetz:
(Might want to update description.)

PR #62, opened 2022-03-11
Codegen TypeScript types for Apollo queries

trm217:
## Description

This PR adds the utility to the www-app to generate TypeScript definitions for all queries that have been defined using the `gql` function imported from the `@apollo/client` package.

The used setup is based on the following blog post:
https://www.apollographql.com/blog/tooling/apollo-codegen/typescript-graphql-code-generator-generate-graphql-types/

Since our code contained queries that didn't have unique names, the naming had to be adapted.


trm217:
Screenshot 2022-03-17 at 14 13 32 2
Due to a non traceable error (shown above) in our graphql queries, getting codegen to work will be a serious time-sink.

PR #64, opened 2022-03-14
feat(access): "mid term" transactional for access grants

fuenkchen:
Context: We want to better accompany our users during access grants. So we add a new transactional mail which users receive _during_ their access grant and which includes article and format recommendations and further information about our magazine.

So this PR 
- adds a new transactional `recommendations`
- adds a new column to `accessCampaigns` table indicating after which duration this recommendations mail should be sent
- adds a new column to `accessGrants` table indicating the timestamp when this recommendations mail was sent
- adds method to `AccessScheduler` to find all suitable grants and activate recommendations mail processing


packages/backend-modules/access/lib/grants.js
```diff
@@ -589,6 +625,33 @@ const beginGrant = async (grant, payload, recipient, pgdb) => {
...
+    beginBefore.subtract(campaign.emailRecommendations[unit], unit),
+  )
+
+  debug('findEmptyRecommendations', { campaign: campaign.id })
+
+  const grants = await pgdb.public.accessGrants.find({
+    accessCampaignId: campaign.id,
+    'beginAt <': beginBefore,
+    recommendationsAt: null,
+  })
```
patrickvenetz, line 641:
```suggestion
  const grants = await pgdb.public.accessGrants.find({
    accessCampaignId: campaign.id,
    'beginAt <': beginBefore,
    invalidatedAt: null,
    recommendationsAt: null,
  })
```

Would add `invalidatedAt: null` here to return valid grants, thus sending recommendations only to users which do have access.

patrickvenetz (approved):
LGTM

PR #65, opened 2022-03-15
Paynote enhance

annatraussnig:
Dropdown in Publikator (tested)
Screen Shot 2022-03-15 at 17 14 28

Special inherited notes on format (tested)
Screen Shot 2022-03-15 at 17 17 27

To do

- [ ] filter out inherited notes when querying format doc itself
- [ ] fix bug where notes stay hidden forever when logged in


packages/backend-modules/documents/graphql/schema-types.js
```diff
@@ -103,6 +108,9 @@ type Meta {
   indicateChart: Boolean
   indicateGallery: Boolean
   indicateVideo: Boolean
+  
+  paynotes: [JSON]
+  tryToBuyRatio: Float
```
tpreusse, line 113: The values we actually return are `null`, `1` and `0` right? Maybe it would be better for have a `paynoteMode` as enum with e.g. `auto`, `try`, `buy` and `none`?

tpreusse (commented):
I like how minimally invasive the code is, but I think it would be better to have a `paynoteMode` instead of `tryToBuyRatio` on the `Meta` object, see in schema types. Would probably also make the publikator code simpler.

packages/backend-modules/documents/graphql/schema-types.js
```diff
@@ -48,6 +48,11 @@ type Newsletter {
   free: Boolean
 }
 
+type PaynoteTarget {
+  hasActiveMembership: Boolean
+  isEligibleForTrial: Boolean
+}
```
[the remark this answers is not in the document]
annatraussnig: 🤦

packages/backend-modules/documents/graphql/schema-types.js
annatraussnig: Good tip. I went for a `paynoteCta` construct, which does clarify the code somewhat.

apps/www/components/Article/Page.js
```diff
@@ -510,10 +510,15 @@ const ArticlePage = ({
...
+              tryToBuyRatio={
+                meta.tryToBuyRatio !== null
+                  ? meta.tryToBuyRatio
+                  : TRY_TO_BUY_RATIO
+              }
               tryOrBuy={payNoteTryOrBuy}
               documentId={documentId}
               repoId={repoId}
-              customPayNotes={article?.content?.meta?.paynotes}
+              customPayNotes={meta.paynotes !== null ? meta.paynotes : []}
```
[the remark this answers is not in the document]
annatraussnig: didn't know `??` 😄

annatraussnig:
Thanks for this first round. I did some refactoring based on your suggestion to replace `tryToBuyRatio` with something more human readable. 

Also included: a störer at the bottom of the profile page (discussed with Katharina).

apps/www/components/Profile/Page.js
```diff
@@ -623,6 +632,17 @@ const LoadedProfile = (props) => {
           <div style={{ clear: 'both' }} />
         </div>
       </MainContainer>
+      {!me && (
+        <PaynoteBanner
+          payNote={paynote.after}
+          overwriteContent={paynoteContent}
```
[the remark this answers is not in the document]
annatraussnig: Indeed – initially it was a buy button and it felt simpler to do it with the generator function. But trialForm cta changes things.

annatraussnig:
Thanks for the suggestion re- profile störer. It does look better:

Screen Shot 2022-04-12 at 17 59 10

I removed the debattenbeiträge in the case where the störer comes up – they looked a bit awkward down there. Let me know what you think.

apps/publikator/components/editor/modules/meta/PaynotesForm.js
```diff
@@ -50,53 +77,96 @@ const DEFAULT_PAYNOTE = {
...
+        )
+      })}
+    </>
+  ),
+)
+
+export default withT(({ t, editor, node, isFormat }) => {
   const paynotes = node.data.get(PAYNOTE_KEY) || []
+  const paynotesCta = node.data.get(PAYNOTE_CTA_KEY)
+  const [mode, setMode] = useState()
```
tpreusse, line 141: ### one more thing

this `useState` and the `useEffect` below should be refactored away and directly inferred from the cta:
```
const mode = paynotesCta || paynotes?.length ? MODE_KEYS.CUSTOM : MODE_KEYS.AUTO
```
(assuming that none is set explicitly in the future – pending comment from earlier)

tpreusse (commented):
awesome, I like the new «native look» pay note look and think it's right to hide the comment count in that case.

annatraussnig:
Yeah that part's not done yet ;)

apps/www/components/Article/PayNote.js
```diff
@@ -244,10 +237,10 @@ const BuyButton = ({ payNote, payload }) => {
         () => router.push(payNote.button.link),
       )}
     >
-      {payNote.button.label}
+      {payNote.button.label || t('article/payNote/default/button')}
```
tpreusse, line 240: rm default

tpreusse (approved):
good for me, two more mini things

PR #68, opened 2022-03-21
7-Stats

tpreusse:
Two scripts I've used for our next «7-Stats» article about our business:

- `matomo/script/visits.js` has been used in the past two years too, it only needed a few [referrer updates](https://github.com/republik/plattform/commit/abf6662c6c72d7a3272108218de6bce4a307e937)
- `republik/script/prolong/analyse.js` is new, it loads all membership periods into memory and finds anniversaries (1 year, 2 years, etc.), expirations (with and without cancelations), changing intervals (up and downgrades) and reactivations (when someones comes back). This data can then be segmented by membership age and membership interval.

Here a preview what I then do with the prolong data:
Screenshot 2022-03-21 at 11 10 00


patrickvenetz (approved):
A masterpiece

PR #76, opened 2022-03-30
Use codemod to get rid off unnecessary `import React from 'react'` imports

trm217:
## Description

Since React 17, importing React in every file is no longer required. With this PR I activated the `'react/react-in-jsx-scope': 'off',` rule in the eslint-config-frontend repo. I then ran the `Removing unused React import`-codemod (referenced in the [React 17 update guide](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports)) across all files (once for all .js files and once for all .tsx files).

## Where Scripts was run:
- www (react 17 transform provided by Next.js)
- publikator (react 17 transform provided by Next.js 9.5.3, we use 9.5.4 - close call)
- admin (react 17 transform provided by Next.js 9.5.3, we use 9.5.5)

## Where Script was not run:
- styleguide (since we no longer use babel, we add the jsx-transformer. Also docs would have to be reac-scripts 4.0 or higher -> importing React is still required)


apps/admin/components/Auth/Poller.js
```diff
@@ -1,4 +1,4 @@
-import React, { Component, Fragment } from 'react'
+import { Component, Fragment } from 'react';
```
tpreusse, line 1: The code mod added semicolons – please re-run it on the current main on Wednesday without semicolons or remove them afterwards by running it all through prettier.

tpreusse (changes requested):

PR #77, opened 2022-03-30
Fix turborepo pipelines

trm217:
## Description

Make sure both self and dependency build tasks are required for test tasks. Bump Turborepo to 1.2.5


turbo.json
```diff
@@ -37,26 +37,6 @@
...
-      "cache": false
-    },
-    "@orbiting/admin-app#build": {
-      "cache": false
-    },
-    "@project-r/styleguide#build": {
-      "cache": false,
-      "dependsOn": [
-        "^build",
-        "$SERVER"
```
tpreusse, line 53: `$SERVER` müsste eigentlich drin bleiben, styleguide build task:
```
yarn build:lib && if [ \"$SERVER\" = \"styleguide\" ]; then yarn build:docs; fi
```

tpreusse (approved):
lgtm

PR #78, opened 2022-03-30
Artikel-Ende: recommendations in Meta

patrickvenetz:
[Notion](https://www.notion.so/republikmagazin/Artikel-End-Lesen-Sie-auch-via-Metadata-969b5fe2d7814bb8bfe8c81a53606275)

**Artikel-Ende before and after**

Ending before and after

**Artikel-Ende before and after (mobile)**

Ending before and after (mobile)

**Publikator UI**

Meta-Block

**Publikator Bot migrating data**

Publizierte Version

**apps/api changes**

Schema type extended:

```gql
type Meta {
  recommendations: DocumentConnection
}
```

It will contain resolved documents linked as repo URLs in `Commit.meta.recommendations[]` e.g.

```
recommendations:
  - 'https://github.com/republik/article-der-russische-schizofaschismus'
  - 'https://github.com/republik/article-zwischen-kabul-und-kiew'
  - 'https://github.com/republik/article-der-streit-um-die-erinnerung'
```

To resolve repo URLs to documents, a multitude of [code parts were enhanced](https://github.com/republik/plattform/pull/78/files#diff-4943c488b8e95194086d70466c90b42cac35db97259872a163ee3d3bbf0a2233R360-R362) (and partially refactored) of which [documents/lib/resolve.js](https://github.com/republik/plattform/pull/78/files#diff-6e140f50376cf59896be10128c61f2841c53546bfb3a806528499ba31defc69aR338-R357) profits aswell as [publikator/graphql/resolvers/Meta.js](https://github.com/republik/plattform/pull/78/files#diff-5f2f2bdce6e525c47f4b7f7d947404ebb624a91d7c7ccf3164b549b8cf3d3dbcR109-R153).

A script will extract "Artikelsammlungen" and add them as `Commit.meta.recommendations` (it will commit and maybe published these changes as "Publikator Bot").


annatraussnig (commented):
This looks so clean 😻 I tested the feature: FE looks good. 

I've got one question (evt. request) regarding Publikator: could we use the `TeaserFeed` component from the Styleguide instead of the custom rendering we've got presently? I guess the "planned publication" and error messages aren't currently supported by the teaser but it shouldn't be possible to add support for this, no?

It doesn't feel quite right to me that we have code that parses and renders the author line in Publikator now. Also then we'd have all the correct title fonts and the format zeile for (almost) free.

packages/backend-modules/publikator/script/extractRecommendations.ts
```diff
@@ -0,0 +1,341 @@
...
+    })
+    debug('publication: %s', publication?.id)
+
+    const isPublished = !!publication
+    debug('isPublished: %o', isPublished)
+
+    const scheduledPublication = await pgdb.publikator.milestones.findOne({
+      repoId,
+      scope: 'publication',
+      scheduledAt: null,
```
fuenkchen, line 62: this confuses me, should `scheduledAt` not be a non-null value?

trm217:
> This looks so clean 😻 I tested the feature: FE looks good.
> 
> I've got one question (evt. request) regarding Publikator: could we use the `TeaserFeed` component from the Styleguide instead of the custom rendering we've got presently? I guess the "planned publication" and error messages aren't currently supported by the teaser but it shouldn't be possible to add support for this, no?
> 
> It doesn't feel quite right to me that we have code that parses and renders the author line in Publikator now. Also then we'd have all the correct title fonts and the format zeile for (almost) free.

@annatraussnig I removed all custom logic and made use of the TeaserFeed component. I added two new props `dense` (to remove the vertical padding) and `nonInteractive` (prevent rendering links) for a nice integration.
image

apps/www/components/Article/ArticleRecommendationsFeed.tsx
```diff
@@ -0,0 +1,57 @@
...
+    },
+  })
+
+  return (
+    <Center>
+      <Loader
+        loading={loading}
+        render={() => (
+          <>
+            {data.article.meta.recommendations?.nodes.length > 0 && (
```
tpreusse, line 37: `data.article` will be null when the article is not found (e.g. path changed in the mean time). Also according to error logs `data` is sometimes `undefined`:
Screenshot 2022-04-22 at 23 23 33

Why are we doing a separate query instead of adding it to the article page query?

trm217: It's a separate query since we don't want it to be included in the SSG version of the page.
The ArticleRecommendationsFeed is only rendered in case a user is authenticated and an active member.

tpreusse: Thank you for the fix. Go for deploying it. I think data being undefined might be in case of error, I think we should display the error (fwd to loader) and move the heading outside the loader, but that's something for next week.

PR #80, opened 2022-03-31
fix(SyntheticAudio): link and removal of test

ovbm:
Screenshot 2022-03-31 at 09 43 54


patrickvenetz:
I will add API change here too, so not only tester but everyone will see synthetic read aloud audio files.

ovbm:
perfect!

patrickvenetz (approved):
LGTM

PR #81, opened 2022-03-31
share chart with one more campaign, update teilen page

fuenkchen:
context: Abo teilen campaign will change from special 8 weeks campaign to a new ordinary 21 days campaign
- update `/teilen` page title, description and share image
- add new campaign id to `ShareChart` component


patrickvenetz (approved):
LGTM

PR #84, opened 2022-04-04
Limit pending sign-in tokens

patrickvenetz:
This Pull Request address [a report from our bug bounty program](https://app.gobugfree.com/account/programs/republik/submissions/KAYPWRGlBq) which could have triggered a vast amount of sign-in emails.

It limits `mutation signIn` with type `EMAIL_TOKEN` to 5 per hour (this imitating an existing rate limit with type `EMAIL_CODE`). It uses a sliding window to open up slots to request more sign-in emails.

It furthermore limits type `APP` – authorizing via App – to 5 requests per 10 minutes.

Once a limit is reached, a `TokensExceededError` is thrown a human readable error message.

It ever so slightly refactors token counting: count tokens instead of fetching them if payloads are not required to generate a new token.

**A screenshot with a sample error message**

Bildschirmfoto 2022-04-05 um 15 56 46

### DIFF
--- a/packages/backend-modules/auth/lib/challenges/AccessTokenChallenge.js
+++ b/packages/backend-modules/auth/lib/challenges/AccessTokenChallenge.js
@@ -21,12 +21,19 @@ module.exports = {
       throw new AccessTokenMissingError({ email })
     }
 
+    // Find tokens matching Type, email and payload (and ignoring expiry date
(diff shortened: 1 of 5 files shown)


fuenkchen (approved):
lgtm 👏

PR #91, opened 2022-04-13
feat(mail): restore interests on Mailchimp if last event was type "cleaned"

patrickvenetz:
This Pull Request checks on MailChimp Webhook event type "subscribe" if latest record in `mailchimLog` for an email address is type "cleaned". If so, it will call `enforceSubscription` and thus updating and restoring interests.

It should resolve a race condition (we hot fixed in 57701d787eb6f2075af936451033cbd41f90d933).


fuenkchen (approved):
👏

PR #93, opened 2022-04-13
feat(onboarding): add dialog hints to onboarding end

ovbm:
Screenshot 2022-04-13 at 14 55 42

Minimal version. Would be nice to have it more integrated in the flow but it doesn't coneptually fit in the "Einrichten" context. All of the current sections have actions attached to them. The text about dialog is more of a hint, and there is no onboarding action to be taken. Also, it is kind of weird to direct people to the comments section before they have a chance to read the magazine. 

In general, I'm not convinced that the dialog needs to be present in the onboarding - except for dialog notification options.

### DIFF
--- a/apps/www/components/Onboarding/Page.js
+++ b/apps/www/components/Onboarding/Page.js
@@ -327,6 +327,33 @@ class Page extends Component {
                 </P>
 
                 <P {...styles.p}>
+                  {t.elements('Onboarding/Sections/Profile/dialogParagraph', {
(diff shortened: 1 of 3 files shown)


patrickvenetz:
Trial users can't participate in debates. Is this notice visible to trial users, too?

ovbm:
Good point. It shouldn't be.

ovbm:
Discussed that this is pointless :D

PR #96, opened 2022-04-25
Fix/read aloud return estimated reading minutes

patrickvenetz:
This Pull Requests moves reading and media time logic to API.

- app/www depends on `Meta.estimatedConsumptionMinutes` solely
- app/api provides either reading or media consumption time. If an `audioSource` is flagged as `readAloud` or `syntheticReadAloud`, it will return estimated reading time, since "read aloud" is a different representation of written content (like a PDF file)
- app/api loses unused `Meta.totalMediaMinutes`

### DIFF
--- a/apps/www/components/ActionBar/index.js
+++ b/apps/www/components/ActionBar/index.js
@@ -151,13 +151,9 @@ const ActionBar = ({
     meta.path,
   )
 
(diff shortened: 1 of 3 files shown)


tpreusse (approved):
lgtm

PR #101, opened 2022-04-28
fix(admin): render address name in payments list

patrickvenetz:
This Pull Request will render `Address.name` in Admin-Tool payments list if string differs from a users name. It helps Finance to find payments easier.

Bildschirmfoto 2022-04-28 um 17 32 38


fuenkchen (approved):
🎉

PR #102, opened 2022-04-29
SSG for front & marketing-page 🏎️ 💨 

trm217:
# Description 📄 

Using the new Next.js Middleware API (as of 1.12.6 still in beta) we now conditionally rewrite users to either the marketing-page ("/") or the front ("/front"). The check on where to rewrite the user to, is done using a newly added JWT, which includes the users email and roles.

The JWT is issued in an express-middleware on the API, which automatically attaches a http-only cookie containing the JWT (signed using a private-key) to the response if the request was made by a logged in user. 
In the frontend this JWT is verified using the matching public-key and the rewriting is then done based on if a user has the member role.

The front and marketing-page, which previously were rendered on the same page, were extracted into separate pages.
Both of these pages utilise SSG with ISR. The new API-key feature provided with #109 was used for the static generation of the front, which contains document data,that is only available to users with an active-membership.

## Steps for deployment ✅ 

1. Ensure that #109 has been merged. (Adds API-key functionality)
2. Add the following env-variables to the production API.
   - JWT_ISSUER = domain // republik.ch
   - JWT_PRIVATE_KEY = b64 encoded private key generated with scripts/generate-keypair.sh
   - JWT_COOKIE_NAME = name of the cookie containing the JWT
   - DOCUMENTS_API_KEYS // In case not already defined
3. Add the following env-variables to the production frontend
   - COOKIE_NAME = // match the env-variable given in the production API
   - JWT_ISSUER = domain // republik.ch
   - JWT_PUBLIC_KEY = b64 encoded public key generated with scripts/generate-keypair.sh
   - JWT_COOKIE_NAME = name of the cookie containing the JWT
   - SSG_DOCUMENTS_API_KEY = one of the keys defined on the API
4. Deploy API
5. Deploy www
6. Be amazed by the sheer speed of the front & marketing-page 🏎️ 💨 

# Important ⚠️ 

Since we make use of the beta-middleware, the Next.js dependency was set to an absolut version (1.12.6), since the middleware API might change with every release until it's out of beta.
Therefor be sure to carefully read the changes made in each new next.js version that is manually bumped.


apps/www/.env.example
```diff
@@ -5,11 +5,22 @@ ASSETS_SERVER_BASE_URL=http://localhost:5020
...
 # Used for external refs and enforing correct protocol and hostname
 PUBLIC_BASE_URL=http://localhost:3010
+NEXTJS_SERVER_HOSTNAME=localhost
 
 # Used for links to CMS and CRM
 PUBLIKATOR_BASE_URL=http://localhost:3005
 ADMIN_BASE_URL=http://localhost:3006
 
+# API key to access restricted GraphQL queries
+# SSG_API_KEY=
```
tpreusse, line 15:
```suggestion
# API key to access restricted document lists via GraphQL queries
# SSG_DOCUMENTS_API_KEY=
```

apps/www/components/Auth/SignOut.js
```diff
@@ -42,7 +43,8 @@ class SignOut extends Component {
                   // re-load after sign in
                   // - clear apollo cache
                   // - clears potentially authenticated ws conntection
-                  window.location.reload()
+                  //window.location.reload()
+                  router.reload()
```
tpreusse, line 47: why change this? seems to do 100% the same thing:
https://github.com/vercel/next.js/blob/d67baa0d1fef8a411679c9b9bdea39cb476ce2d5/packages/next/shared/lib/router/router.ts#L879-L881
https://nextjs.org/docs/api-reference/next/router#routerreload

apps/www/components/Frame/index.js
```diff
@@ -104,16 +101,16 @@ const Frame = ({
...
   isOnMarketingPage,
   pageColorSchemeKey,
 }) => {
   const { inNativeApp, inNativeAppLegacy } = useInNativeApp()
-  const { me } = useMe()
+  const { t } = useTranslation()
+  const { me, hasActiveMembership } = useMe()
 
-  const hasOverviewNav = isMember && wantOverviewNav
+  const hasOverviewNav = hasActiveMembership && wantOverviewNav
```
tpreusse, line 113: `hasActiveMembership` is not the right thing here, should be `hasAccess` instead (isMember = has role member = hasAccess – includes access grants, hasActiveMembership = is paid member)

apps/www/lib/context/MeContext.tsx
```diff
@@ -88,16 +98,60 @@ export const useMe = (): MeContextValues => useContext(MeContext)
 
 type Props = {
   children: ReactNode
+  /**
+   * Assumes that a memberships exists, even before me is loaded.
+   * All values returned from the context that correlate to a membership will be set to true.
+   */
+  assumeMembership?: boolean
```
tpreusse, line 105: `assumeAccess` is probably what we need

apps/www/lib/context/MeContext.tsx
```diff
@@ -42,7 +50,9 @@ css.global(
   },
 )
 
-type Me = {
+const AUTH_STATE_CHANGE_BROADCAST_STORAGE_KEY = 'authStateChangeBroadcast'
```
tpreusse, line 53: couldn't we just listen to `ME_PORTRAIT_STORAGE_KEY`?

apps/www/pages/_middleware.tsx
```diff
@@ -0,0 +1,79 @@
...
+      isInNativeIOSApp &&
+      !NativeAppHelpers.isNewerVersion(
+        '2.1.0',
+        NativeAppHelpers.getNativeAppVersion(userAgent),
+      )
+    ) {
+      resUrl.pathname = '/anmelden'
+    } else {
+      resUrl.pathname = '/'
+    }
```
tpreusse, line 60: this can actually be removed, was only necessary before apple approved the new marketing page with trial (Probelesen)

apps/www/pages/_middleware.tsx
```diff
@@ -0,0 +1,79 @@
...
+          },
+          method: 'POST',
+          body: err.stack,
+        })
+          .catch((err) =>
+            console.log('Failed to report error in middleware', err, resUrl),
+          )
+          .finally(() => resolve(null))
+      }),
+    )
```
tpreusse, line 77: why not just `console.error` here? This always runs on the server and could just do a simple log instead of making an http request that then ends up making a simple log ;-)

apps/www/.env.example
```diff
@@ -5,11 +5,22 @@ ASSETS_SERVER_BASE_URL=http://localhost:5020
 
 # Used for external refs and enforing correct protocol and hostname
 PUBLIC_BASE_URL=http://localhost:3010
+NEXTJS_SERVER_HOSTNAME=localhost
```
tpreusse, line 8: IMO just compute this:
```
new URL(PUBLIC_BASE_URL).hostname
```

packages/backend-modules/auth/express/auth.js
```diff
@@ -56,8 +63,8 @@ exports.configure = ({
       cookie: {
         domain,
         maxAge: maxAge,
-        sameSite: !dev && 'none',
-        secure: !dev,
+        sameSite: cookieOptions.sameSite,
+        secure: cookieOptions.secure,
```
tpreusse, line 67: Wouldn't it make more sense to spread `cookieOptions` and always do domain via there?

packages/backend-modules/auth/lib/jwt.js
```diff
@@ -0,0 +1,45 @@
+const jwt = require('jsonwebtoken')
+const { userIsInRoles, specialRoles } = require('./Roles')
+
+// TODO: copy of code in express/auth.js move these vars to constants-file
```
tpreusse, line 4: 😉

apps/www/lib/context/MeContext.tsx
```diff
@@ -135,6 +189,25 @@ const MeContextProvider = ({ children }: Props) => {
...
+    }
+  }, [getAuthState])
+
+  // Register a storage-event listener
+  useEffect(() => {
+    window.addEventListener('storage', handleAuthenticationChangeBroadcast)
+    return () => {
+      window.removeEventListener('storage', handleAuthenticationChangeBroadcast)
+    }
+  }, [handleAuthenticationChangeBroadcast])
```
tpreusse, line 209: currently local storage calls are made outside try catchs in `broadcastAuthenticationChange` which is unsafe and will explode in some browsers.

Probably better of using our existing `createPersistedState` which already does cross tab synch, ideally with `ME_PORTRAIT_STORAGE_KEY`.

packages/backend-modules/auth/express/auth.js
trm217: Probably yes, however I don't know what the motivation was behind passing all arguments to auth.js its self. I'd like Pae's opinion on that.

apps/www/lib/context/MeContext.tsx
trm217: In regards to having issues in some browsers: Who ever uses a browser that doesn't support [localStorage](https://caniuse.com/?search=localstorage) in 2022 or has somehow messed up his browser-config is totally deserving off the site crashing on him ^^

Using ME_PORTRAIT_STORAGE_KEY seems very wrong to me. Especially since it means that we would effectively duplicate the portrait-state of the logged in user within the context.

What about creating a hash from the user object and sharing that?
This way we could also ensure that the sync only happens if the Me object has effectively changed.

apps/www/pages/_middleware.tsx
trm217: I shall agree but still be unhappy that we don't use a nice error logging service :P

apps/www/lib/context/MeContext.tsx
trm217: Let's discuss this on the other comment

apps/www/pages/_middleware.tsx
trm217: I disagree. While errors in the middleware are most likely to occur due to "server"-side errors it still might be causes due to the request (especially it's headers). In fact, I think we should also log the jwt cookie in the error as-well as the presence of the session cookie. And I still think we should parse the user agent. Non consistent logs are very undesirable. Also the strain on the server is negligible (also it doesn't slow down the middleware since the log to the api-route is made after the middle-ware response has been returned to the user.)

tpreusse: Logging (session) cookies is a absolute no go. Also isn't this an bad loop where the request passes through the middleware again? Or are api request excluded from the middleware?

trm217: Not the session cookie, just the JWT (not usable without the session-cookie anyways). And since it's a post request with HTTPS it would also not be accesible.
Every request goes through the middleware, however this isn't problematic since we have a guard clause that immediatly return if the route is != '/'

trm217: Not the session cookie, just the JWT (not usable without the session-cookie anyways). And since it's a post request with HTTPS it would also not be accesible.
Every request goes through the middleware, however this isn't problematic since we have a guard clause that immediatly return if the route is != '/'

trm217: imagine something like:
```
{
  sessionCookiePresent: true,
  jwt: "dasfsdf.adsfafdsfdsaadfs.afsdfdas"
}
```
This would provide us all information we need to see why rewrite might have failed.

apps/www/server/index.js
```diff
@@ -20,6 +20,10 @@ const { CURTAIN_MESSAGE } = process.env
 
 const app = next({
   dev: DEV,
+  port: PORT,
+  hostname: process.env.PUBLIC_BASE_URL
+    ? new URL(process.env.PUBLIC_BASE_URL).hostname
+    : 'localhost',
```
tpreusse, line 26: maybe it would be better to fall back to `undefined`? Or even refuse to start the server without `PUBLIC_BASE_URL` since it's necessary?

apps/www/pages/_middleware.tsx
tpreusse: Honestly I don't get see the point of this whole try catch block – does next docu suggest doing this? Let's discuss later over voice

apps/www/server/index.js
trm217: Good catch! 👍🏻

apps/www/pages/_middleware.tsx
trm217: There are countless errors that can occure in the above code, for example during JWT verification/validation.
The main reason why I want to log the errors is that if the middleware pages, both the front-breaks.

apps/www/pages/_middleware.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const jwtBody = await parseAndVerifyJWT(req)
+    const isMember = jwtBody?.roles?.includes('member')
+
+    if (isMember) {
+      resUrl.pathname = '/front'
+      return NextResponse.rewrite(resUrl)
+    }
+  } catch (err) {
+    reportError(null, 'JWT Verification Error')
+    console.error(err)
```
tpreusse, line 45: Sonst wird ja `reportError` geladen, was `useragent` lädt, was ja scheinbar ned geht ;)
```suggestion
    console.error('JWT Verification Error', err)
```

apps/www/pages/_ssr/front-preview/[...id].tsx
```diff
@@ -0,0 +1,61 @@
...
+    return {
+      props: {
+        id: params.id,
+      },
+    }
+  } catch (error) {
+    reportError(ctx.req.headers['user-agent'], error)
+    return {
+      notFound: true,
+    }
```
Comment (tpreusse) [inline apps/www/pages/_ssr/front-preview/[...id].tsx:59]: rm whole try catch here – it better and more visible if it explodes if it doesn't work

apps/www/pages/_middleware.tsx
trm217: Vögel!

apps/www/lib/context/MeContext.tsx
trm217: removed as discussed

apps/www/pages/_middleware.tsx
trm217: Moved try-catch arround jwt verification instead of arround the entire middleware

apps/www/components/Front/index.js
```diff
@@ -161,6 +134,34 @@ const Front = ({
...
+            children: {
+              ...previousResult.front.children,
+              nodes: [...previousNodes, ...currentNodes],
+              pageInfo: currentSearch.pageInfo,
+            },
+          },
+        }
+        return res
+      },
+    }).catch((error) => console.error(error))
```
tpreusse, line 162: if we do not catch it our report error with catch it as uncaught exception
```suggestion
    })
```

apps/www/lib/context/MeContext.tsx
```diff
@@ -88,11 +96,27 @@ export const useMe = (): MeContextValues => useContext(MeContext)
...
   const { data, loading, error, refetch } = useQuery<MeResponse>(meQuery, {})
 
+  const getAuthState = useCallback((): AuthState => {
+    if (data && data.me) {
+      return 'logged-in'
+    }
+    return 'logged-out'
+  }, [data])
+
+  const [authState, setAuthState] = useState(getAuthState())
```
tpreusse, line 118:
```suggestion
```

apps/www/lib/context/MeContext.tsx
```diff
@@ -88,11 +96,27 @@ export const useMe = (): MeContextValues => useContext(MeContext)
...
   children: ReactNode
+  /**
+   * Assumes that a memberships exists, even before me is loaded.
+   * All values returned from the context that correlate to a membership will be set to true.
+   */
+  assumeAccess?: boolean
 }
 
-const MeContextProvider = ({ children }: Props) => {
+type AuthState = 'loading' | 'logged-in' | 'logged-out'
```
tpreusse, line 106:
```suggestion
```

apps/www/pages/front.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const { me, meLoading, hasActiveMembership } = useMe()
+
+  useEffect(() => {
+    if (meLoading) return
+    // In case the user isn't logged in or has no active membership,
+    // reload to re-trigger the middleware to rewrite to the marketing-page
+    if (!me || !hasActiveMembership) {
+      router.reload()
+    }
+  }, [me, meLoading, hasActiveMembership, router])
```
tpreusse, line 26:
```suggestion
  const { meLoading, hasAccess } = useMe()

  useEffect(() => {
    if (meLoading) return
    // reload to re-trigger the middleware to rewrite to the marketing-page
    if (!hasAccess) {
      window.location.reload()
    }
  }, [meLoading, hasAccess])
```

apps/www/pages/index.js
```diff
@@ -1,39 +1,31 @@
...
+  const { me, meLoading, hasActiveMembership } = useMe()
+
+  useEffect(() => {
+    if (meLoading) return
+    // In case the user is logged in and has an active membership,
+    // reload the page to rewrite from '/' to '/front'
+    if (me && hasActiveMembership) {
+      router.reload()
+    }
+  }, [me, meLoading, hasActiveMembership, router])
```
tpreusse, line 28:
```suggestion
  const { meLoading, hasAccess } = useMe()

  useEffect(() => {
    if (meLoading) return
    // reload the page to rewrite from '/' to '/front' via middleware
    if (hasAccess) {
      window.location.reload()
    }
  }, [meLoading, hasAccess])
```

apps/www/server/index.js
```diff
@@ -18,8 +16,18 @@ const PORT = process.env.PORT || 3005
 
 const { CURTAIN_MESSAGE } = process.env
```
tpreusse, line 17:
```suggestion
const { CURTAIN_MESSAGE, PUBLIC_BASE_URL } = process.env
```

apps/www/server/index.js
```diff
@@ -18,8 +16,18 @@ const PORT = process.env.PORT || 3005
...
 
+if (!process.env.PUBLIC_BASE_URL) {
```
tpreusse, line 19:
```suggestion
if (!PUBLIC_BASE_URL) {
```

apps/www/server/index.js
tpreusse, line 17:
```suggestion
const { CURTAIN_MESSAGE, PUBLIC_BASE_URL } = process.env
```

apps/www/server/index.js
```diff
@@ -18,8 +16,18 @@ const PORT = process.env.PORT || 3005
...
+    'missing PUBLIC_BASE_URL environment variable, but is required by next-js middleware.',
+  )
+}
+
 const app = next({
   dev: DEV,
+  port: PORT,
+  hostname: process.env.PUBLIC_BASE_URL
+    ? new URL(process.env.PUBLIC_BASE_URL).hostname
+    : undefined,
```
tpreusse, line 30: Since we now never get here without a PUBLIC_BASE_URL we can just so this:
```suggestion
  hostname: new URL(PUBLIC_BASE_URL).hostname
```

apps/www/lib/context/MeContext.tsx
trm217: Upsala :s

apps/www/lib/helpers/createGetStaticProps.ts
```diff
@@ -26,19 +26,31 @@ type ApolloSSGQueryFunc<P, Q extends ParsedUrlQuery> = (
...
+      return result
     }
 
-    return result
+    result.props[APOLLO_STATE_PROP_NAME] = apolloClient.cache.extract()
+
+    return {
+      props: {
+        ...result.props,
+        [APOLLO_STATE_PROP_NAME]: apolloClient.cache.extract(),
```
tpreusse, line 49: `[APOLLO_STATE_PROP_NAME]: apolloClient.cache.extract()` is doublicated

packages/backend-modules/auth/lib/Sessions.js
```diff
@@ -12,12 +13,20 @@ const InitiateSessionError = newAuthError(
...
       if (error) {
         return reject(new DestroySessionError({ headers: req.headers, error }))
       }
+      if (res) {
+        const cookieOptions = getCookieOptions()
+        res.clearCookie(process.env.COOKIE_NAME ?? 'connect.sid', cookieOptions) // clear session cookie
+        res.clearCookie(
+          process.env.JWT_COOKIE_NAME ?? 'republik-token',
+          cookieOptions,
+        ) // clear jwt cookie
```
tpreusse, line 28:
```suggestion
        res.clearCookie(process.env.COOKIE_NAME ?? 'connect.sid', cookieOptions)
        res.clearCookie(
          process.env.JWT_COOKIE_NAME ?? 'republik-token',
          cookieOptions,
        )
```
what information do those code comments add?

apps/www/pages/_ssr/front-preview/[...id].tsx
```diff
@@ -0,0 +1,53 @@
+import { GetServerSideProps, NextPage } from 'next'
```
Comment (tpreusse) [inline apps/www/pages/_ssr/front-preview/[...id].tsx:1]: AFAICS we don't need to accept multiple path segments here and `id` seems like a front id but is then mapped to `extractId`. Also preview is confusing with the `vorschau` path that we also support.

```diff
-apps/www/pages/_ssr/front-preview/[...id].tsx
+apps/www/pages/_ssr/front/[extractId].tsx
```

apps/www/pages/_ssr/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+import { PUBLIC_BASE_URL, CDN_FRONTEND_BASE_URL } from '../../lib/constants'
+
+import withDefaultSSR from '../../lib/hocs/withDefaultSSR'
+
+const IndexPage = ({ t, isMember, router }) => {
+  const { inNativeIOSApp, isMinimalNativeAppVersion } = useInNativeApp()
+  if (
+    router.query.stale !== 'marketing' &&
+    (isMember || router.query.extractId)
```
tpreusse, line 20: this whole file is never used anymore right? IMO remove

tpreusse (commented):
this PR is a BEAST ;-)

apps/www/pages/_middleware.tsx
```diff
@@ -0,0 +1,55 @@
...
+  try {
+    // Parse and verify JWT to decide about redirection
+    const jwtBody = await parseAndVerifyJWT(req)
+    const isMember = jwtBody?.roles?.includes('member')
+
+    if (isMember) {
+      resUrl.pathname = '/front'
+      return NextResponse.rewrite(resUrl)
+    } else {
+      const sessionCookieString = req.cookies?.[process.env.COOKIE_NAME]
```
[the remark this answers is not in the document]
trm217: Won't work, that's in the backend.

apps/www/pages/index.js
```diff
@@ -1,39 +1,41 @@
...
+  const [syncingUser, setSyncingUser] = useState(false)
+
+  useEffect(() => {
+    if ((router.query.syncUser = '1' && meLoading)) {
+      setSyncingUser(true)
+    } else if (syncingUser && !meLoading) {
+      router.replace(router.pathname, null, { shallow: true })
+      setSyncingUser(false)
+    }
+  }, [router.query, meLoading])
```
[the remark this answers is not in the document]
trm217: Yeah that was hacky as hell ^^
I should've just gone to bed 🛌🏻

apps/www/next.config.js
```diff
@@ -55,11 +55,29 @@ module.exports = withTM(
     },
     async redirects() {
       return [
+        // '/front' must not be reachable directly!
+        // if a user is authenticated, '/' will be rewritten to '/front'
+        {
+          source: '/front',
+          destination: '/',
+          permanent: true,
+        },
```
tpreusse, line 64: `permanent`  might not be ideal since we lock that path if we'd ever want to use it
```suggestion
        {
          source: '/front',
          destination: '/',
        },
```
or even rewrite to 404

tpreusse (approved):

PR #107, opened 2022-05-06
feat(Table): add bar chart option

felixmichel:
This PR adds a bar chart feature to the table.

Bildschirmfoto 2022-05-06 um 10 28 05

### DIFF
--- a/packages/styleguide/src/components/Chart/Table.js
+++ b/packages/styleguide/src/components/Chart/Table.js
@@ -297,11 +297,11 @@ const Cell = (props) => {
     numberFormatter,
   } = props
 
-  const parentWidth = Math.max(width, 100)
+  const maxWidth = Math.max(width, 100)
 
   const barScale = scaleLinear()
     .domain(barChartExtent)
-    .range([5, parentWidth - PADDING])
+    .range([0, maxWidth - PADDING])
 
   return (
     <td
@@ -317,9 +317,8 @@ const Cell = (props) => {
     >
       {isBarColumn ? (
         <BarComponent
-          parentWidth={parentWidth - PADDING}
           colorScale={colorScale}
-          barWidth={barScale(value)}
+          barWidth={Math.ceil(barScale(value))}
           label={numberFormatter(value)}
           columnName={columnName}
           color={color}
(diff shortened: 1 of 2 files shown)


packages/styleguide/src/components/Chart/Table.js
```diff
@@ -254,18 +284,107 @@ Table.propTypes = propTypes
...
+  //   if (node !== null) {
+  //     setParentWidth(node.getBoundingClientRect().width)
+  //   }
+  // })
+
+  const parentWidth = Math.max(width, 100)
+
+  const barScale = scaleLinear()
+    .domain(barChartExtent)
+    .range([5, parentWidth - PADDING])
```
tpreusse, line 311: why start at `5`? Should zero be zero pixel? If you want to ensure `0.1` is at least 1 px IMO you could `Math.ceil` the `barWidth` after the scale.

packages/styleguide/src/components/Chart/Table.js
```diff
@@ -254,18 +284,107 @@ Table.propTypes = propTypes
...
   )
 }
+
+const BarComponent = (props) => {
+  const { parentWidth, barWidth, color, label, columnName, colorScale } = props
+
+  const labelSize = labelGauger(label)
+  const BAR_LABEL_PADDING = 10
+
+  const isLabelOutside = labelSize > +barWidth - BAR_LABEL_PADDING
```
tpreusse, line 348: Note sure if the `BAR_LABEL_PADDING` is needed, and `barWidth` comes from scale – doesn't need casting afais
```suggestion
  const isLabelOutside = labelSize > barWidth
```

packages/styleguide/src/components/Chart/Table.js
```diff
@@ -254,18 +284,107 @@ Table.propTypes = propTypes
...
+  } = props
+
+  // const [parentWidth, setParentWidth] = useState(null)
+  // const parentCell = useCallback((node) => {
+  //   if (node !== null) {
+  //     setParentWidth(node.getBoundingClientRect().width)
+  //   }
+  // })
+
+  const parentWidth = Math.max(width, 100)
```
tpreusse, line 307: clean up and get rid of parent naming

tpreusse (approved):

PR #108, opened 2022-05-06
Extract pgp and reportError handlers into next.js api-routes

trm217:
# Description

Extracted two route-handlers from the www express-server and converted them into Next.js api-routes.


apps/www/pages/api/pgp/[userSlug].ts
```diff
@@ -0,0 +1,60 @@
...
+  async function (req: NextApiRequest, res: NextApiResponse) {
+    const apolloFetch = createApolloFetch({
+      uri: process.env.API_URL,
+    })
+    let { userSlug: userSlugWithSuffix } = req.query
+    if (Array.isArray(userSlugWithSuffix)) {
+      userSlugWithSuffix = userSlugWithSuffix.join('')
+    }
+
+    if (!userSlugWithSuffix || !userSlugWithSuffix?.endsWith('.asc')) {
```
Comment (tpreusse) [inline apps/www/pages/api/pgp/[userSlug].ts:16]:
```suggestion
    if (!userSlugWithSuffix?.endsWith('.asc')) {
```

PR #110, opened 2022-05-09
Expandable link

annatraussnig:
Bare bones framework for expandable links

### Demo

https://user-images.githubusercontent.com/3907984/174611112-2a00c1b1-57a9-4339-a5fa-c7c5a5c47394.mov


packages/styleguide/src/components/ExpandableLink/ExpandableLinkCallout.tsx
```diff
@@ -0,0 +1,236 @@
...
+        {...css({
+          animation: `0.3s ${slideUp} forwards`,
+          bottom: -calloutHeight,
+          paddingBottom: inNativeApp ? '50px' : '24px',
+          [mUp]: {
+            animation: 'none',
+            bottom: 'auto',
+          },
+        })}
+        onClick={(e) => e.stopPropagation()}
```
annatraussnig, line 211: why is this here?

apps/publikator/components/editor/modules/link/index.js
```diff
@@ -10,22 +11,31 @@ export default ({ rule, subModules, TYPE }) => {
...
+        nodes: visitChildren(node),
+      }
+    },
+    toMdast: (object, index, parent, { visitChildren }) => {
+      return {
+        type: 'link',
+        title: object.data.description
+          ? [object.data.title || '', object.data.description].join(SEPARATOR)
+          : object.data.title,
+        description: object.data.description,
```
tpreusse, line 34:
```suggestion
```

apps/publikator/components/editor/modules/link/ui.js
```diff
@@ -222,6 +223,14 @@ export const LinkForm = withT(
                 value={node.data.get('title')}
                 onChange={onInputChange('title')}
               />
+              <Field
+                label={t(`link/description`, undefined, 'title')}
```
tpreusse, line 227:
```suggestion
                label={t(`link/description`, undefined, 'description')}
```

apps/publikator/components/editor/modules/link/index.js
```diff
@@ -10,22 +11,31 @@ export default ({ rule, subModules, TYPE }) => {
...
+        },
+        nodes: visitChildren(node),
+      }
+    },
+    toMdast: (object, index, parent, { visitChildren }) => {
+      return {
+        type: 'link',
+        title: object.data.description
+          ? [object.data.title || '', object.data.description].join(SEPARATOR)
+          : object.data.title,
```
tpreusse, line 33:
```suggestion
        title: [object.data.title, object.data.description].filter(Boolean).join(SEPARATOR),
```

apps/publikator/components/editor/modules/link/ui.js
```diff
@@ -222,6 +223,14 @@ export const LinkForm = withT(
...
+                value={node.data.get('description')}
+                onChange={onInputChange('description')}
+                renderInput={({ ref, ...inputProps }) => (
+                  <AutosizeInput {...inputProps} inputRef={ref} />
```
tpreusse, line 231: Screenshot 2022-07-21 at 01 24 30

needs `{...styles.autoSize}` to look properly (min height and paddings)

packages/styleguide/src/templates/Article/base.js
```diff
@@ -35,20 +35,31 @@ import {
...
 
 const createBase = ({ metaBody, metaHeadlines }) => {
   const link = {
     matchMdast: matchType('link'),
-    props: (node) => ({
-      title: node.title,
-      href: node.url,
-    }),
+    props: (node, index, parent, { ancestors }) => {
+      const allowsExpandable = ancestors.find(matchParagraph)
```
tpreusse, line 45: Unused and afaik always true anyhow (inline `link` always has a `paragraph` somewhere up the tree in mdast afaik)

tpreusse (commented):
I hate allowing html but I think you wanted the quick solution ;-)

Anyhow should we explain in Publikator that html is allowed and give bold and italic example for quick ref? Any other use cases? sup sub? Are those cases tested? Maybe add it to the docs too.

Also hovering or clicking expandable links in publikator currently throws an exception:
Screenshot 2022-07-21 at 01 30 03

packages/styleguide/src/templates/Article/base.js
annatraussnig: maybe I'm getting confused but i think that way you can't do expandable links in captions and such.

tpreusse: if u check the variable: it only occurs 3 times and is never read for actual usage

annatraussnig: yeah ok, you're right…

apps/publikator/components/editor/modules/link/ui.js
annatraussnig: thanks for the tip!

annatraussnig:
- publikator bug fixed
- docs on formatting added

packages/styleguide/src/components/ExpandableLink/ExpandableLink.tsx
```diff
@@ -0,0 +1,148 @@
...
+  link: css({
+    cursor: 'pointer',
+    padding: '0 2px',
+    textDecorationLine: 'underline',
+    textDecorationStyle: 'dotted',
+    textDecorationSkip: 'ink',
+    textDecorationThickness: 2,
+    textUnderlineOffset: 3,
+  }),
+}
```
tpreusse, line 32: the inconsistency triggers me :D one time it's above and one time below the component in this PR

packages/styleguide/src/components/ExpandableLink/ExpandableLinkCallout.tsx
```diff
@@ -0,0 +1,241 @@
...
+  inNativeApp?: boolean
+  timeOutRef: React.MutableRefObject<NodeJS.Timeout>
+  expandedLink: StateProps
+  setExpandedLink: React.Dispatch<React.SetStateAction<StateProps>>
+}
+
+const ExpandableLinkP = ({ children, ...props }) => {
+  const [colorScheme] = useColorContext()
+  return (
+    <p {...props} {...styles.contentText} {...colorScheme.set('color', 'text')}>
```
tpreusse, line 22:
```suggestion
    <p {...props} {...styles.contentText} {...Interaction.fontRule} {...colorScheme.set('color', 'text')}>
```

Currently faux bold and italic is rendered. By adding [`Interaction.fontRule`](https://styleguide.republik.ch/typography#bold-and-cursive) (needs to be imported as `interactionFontRule` in styleguide) we can fix this and also support `b` nested in `i` etc.

packages/styleguide/src/components/ExpandableLink/ExpandableLinkCallout.tsx
```diff
@@ -0,0 +1,241 @@
...
+const fadeIn = keyframes({
+  from: {
+    opacity: 0,
+  },
+  to: {
+    opacity: 1,
+  },
+})
+
+const shortenLink = (href) => {
```
tpreusse, line 37: While I do not attempt to understand this code the suffix of `/.../` does not seem useful to me in the following examples:

Screenshot 2022-07-25 at 18 19 39
Screenshot 2022-07-25 at 18 19 35

Full urls:
https://eatforum.org/eat-lancet-commission/
https://www.quarks.de/umwelt/klimawandel/darum-sollten-wir-ueber-methan-sprechen/

yarn.lock
```diff
@@ -8250,7 +8250,7 @@ color@^3.0.0:
     color-convert "^1.9.3"
     color-string "^1.6.0"
 
-color@^4.0.1:
+color@^4.0.1, color@^4.2.3:
```
tpreusse, line 8253: those changes in the lock file seem odd and unnecessary. Revert from lock file to state in main?

packages/styleguide/src/components/ExpandableLink/ExpandableLink.tsx
```diff
@@ -0,0 +1,148 @@
...
+      {...colorScheme.set('color', 'text')}
+      {...colorScheme.set('textDecorationColor', 'textSoft')}
+      aria-label={description}
+      onClick={(event) => {
+        event.preventDefault()
+        showInfoBox(event)
+      }}
+      onMouseLeave={removeInfoBox}
+      onMouseEnter={(event) => {
+        showInfoBox(event)
```
tpreusse, line 130: on desktop it's currently possible to trigger a callout on hover. IMO this should be prevented and do nothing when showing the mobile callout that is screen filling.

packages/styleguide/src/components/ExpandableLink/ExpandableLink.tsx
```diff
@@ -0,0 +1,148 @@
...
+    <a
+      {...attributes}
+      {...styles.link}
+      {...colorScheme.set('color', 'text')}
+      {...colorScheme.set('textDecorationColor', 'textSoft')}
+      aria-label={description}
+      onClick={(event) => {
+        event.preventDefault()
+        showInfoBox(event)
+      }}
```
tpreusse, line 127: Shouldn't we allow cmd click to directly get to the source?
```suggestion
      href={href}
      onClick={(event) => {
        if (shouldIgnoreClick(event)) {
          return
        }
        event.preventDefault()
        showInfoBox(event)
      }}
```
(`shouldIgnoreClick` function would need to be moved to styleguide)

tpreusse (changes requested):
A few more comment and additionally currently a multi line description can mess up the schema:
Screenshot 2022-07-25 at 18 01 23

We have to see if there is a technical fix by encoding `\n` or if we should disallow it. (@annatraussnig is already informed)

packages/styleguide/src/components/ExpandableLink/ExpandableLinkCallout.tsx
ovbm: We shouldn't support any styling / bold / italic in the callout. It should only render plain text.

packages/styleguide/src/components/ExpandableLink/ExpandableLink.tsx
ovbm: I am blind. In what file is style below the component?

packages/styleguide/src/components/ExpandableLink/ExpandableLinkCallout.tsx
tpreusse: Well it was implemented with html support like we discussed when specing:
[
Screenshot 2022-07-26 at 10 45 59
](https://www.notion.so/republikmagazin/Inline-Link-Meta-Info-a3be338ba0bc4ea29a68187f78bea4f0)

italic and sub are necessary and very common – use for foreign words and carbon dioxide. bold and sup might not strictly be necessary – adjust the docs to only mention italic and sub?

packages/styleguide/src/components/ExpandableLink/ExpandableLink.tsx
tpreusse: In `ExpandableLinkCallout` it's in the middle – below `ExpandableLinkP` (which uses one styles rule) and above `ExpandableLinkCallout` (which uses the rest). Generally it more logical to have it above since you should declare variables before using them.

packages/styleguide/src/components/ExpandableLink/ExpandableLinkCallout.tsx
ovbm: not sure...

annatraussnig:
Thanks for the thorough testing @tpreusse! We decided to disallow newlines, since the design of the feature only has very short texts in mind. but the replace function used makes it quite easy to encode the newlines instead, should ~we~ popular pressure change our minds.

yarn.lock
annatraussnig: gone ⚔️

packages/styleguide/src/components/ExpandableLink/ExpandableLink.tsx
ovbm: done

packages/styleguide/src/components/ExpandableLink/ExpandableLink.tsx
ovbm: I understand the first part but not your suggestion: do you want to show nothing at all when we on desktop, but with a mobile screen-width?

ovbm: got it

ovbm:
Made all requested changes. Ready for final testing from my side. Thanks for review and effort everyone :D.

ovbm (approved):

packages/styleguide/src/templates/Article/base.js
```diff
@@ -35,20 +35,29 @@ import {
...
+    },
     component: (props) => {
-      const { href } = props
+      const { href, description } = props
+      const LinkComponent = description ? ExpandableLink : Editorial.A
       // workaround app issues with hash url by handling them ourselves and preventing the default behaviour
       if (href && href.slice(0, 3) === '#t=') {
         return (
-          <Editorial.A
+          <LinkComponent
```
tpreusse, line 60: `ExpandableLink` does not support custom `onClick` handler (see below). IMO skip supporting `ExpandableLink` for `#t=` links (media time setting)

```suggestion
          <Editorial.A
```

packages/styleguide/src/templates/Article/base.js
```diff
@@ -62,7 +71,7 @@ const createBase = ({ metaBody, metaHeadlines }) => {
       }
       if (href && href[0] === '#') {
         return (
-          <Editorial.A
+          <LinkComponent
```
tpreusse, line 74: see above – `ExpandableLink` does not have hash routing support with a custom on click – should be skipped.

```suggestion
          <Editorial.A
```

tpreusse (approved):
lgtm but IMO we should adapt `templates/Article/base.js` to skip support for hash links with custom onclick handlers – also unclear what the use case would be and what the expected onclick behaviour should be

apps/publikator/lib/translations.json
```diff
@@ -1304,6 +1304,14 @@
       "key": "link/repo/search",
       "value": "Dokument suchen"
     },
+    {
+      "key": "link/description",
+      "value": "Zusätzliche Beschreibung"
+    },
+    {
+      "key": "link/description/help",
+      "value": "Formatting geht mit <a href=\"https://styleguide.republik.ch/expandablelink#formatting\" target=\"_blank\">HTML-Tags</a>."
```
tpreusse, line 1313:
```suggestion
      "value": "<a href=\"https://styleguide.republik.ch/expandablelink#formatting\" target=\"_blank\">Einfache HTML-Tags erlaubt</a>."
```

packages/styleguide/src/components/ExpandableLink/docs.md
```diff
@@ -0,0 +1,29 @@
...
+Renders as follow:
+
+```react
+<Editorial.P><ExpandableLink description="<b>Brand</b> <em>new</em> data on CO<sub>2</sub> levels" href="https://www.republik.ch" t={t}>Link</ExpandableLink> with html formatting.</Editorial.P>
+```
+
+Supported html tags for formatting:
+
+- b tags for *<b>bold</b>*
+- em tags for **<em>italics</em>**
```
tpreusse, line 28: IMO either `strong` and `em` or `b` and `i` (both are supported via font rule css)

tpreusse (commented):
Noch ein mini ding @annatraussnig

annatraussnig:
alles durch @tpreusse

PR #112, opened 2022-05-17
fix(Paynote): fix doubly escaped characters

annatraussnig:
fix "Mehr Informationen zur Republik." link in the Abo-kaufen paynote.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ build
 *.pem
 local/
 .idea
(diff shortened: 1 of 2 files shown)


tpreusse (approved):
lgtm, if you want I can deploy it with the other stuff over the next few hours

annatraussnig:
yeah that's perfect, thank you!!

PR #113, opened 2022-05-17
add functionality to revoke perks on access grants

fuenkchen:
Some perks, like accessWithRole, should not only be added but be revoked too, when grants end and users do not have a membership yet.

So this PR:
* adds revoke method to perks and to grants lib, which will be called when grant gets invalidated
* adds method to remove roles in access grant membership lib

Not part of this change: All grants which have already been invalidated. We should remove special roles from non-member users manually with an SQL statement.


packages/backend-modules/access/lib/memberships.js
```diff
@@ -21,6 +21,22 @@ const addRole = async (grant, user, pgdb, role) => {
   return false
 }
 
+const removeRole = async (grant, user, pgdb, role) => {
```
patrickvenetz, line 24: `addRole` was a helper function for `addMemberRole` and later exposed to avoid code repetation.

Would implement `removeRole` into [`removeMemberRole`](https://github.com/republik/plattform/pull/113/files#diff-86d6bc219d3e87a6d49a689923f741e568d7466771b83993bbb83a3c39edd1c4R62-R75) now, too.

packages/backend-modules/access/lib/grants.js
```diff
@@ -432,13 +465,19 @@ const invalidate = async (grant, reason, t, pgdb, mail) => {
         })
       }
 
+      const perks = await revokePerks(grant, recipient, campaign, pgdb)
+      if (perks.length > 0) {
+        grant.perks = {}
+
+        await Promise.map(perks, (perk) => {
+          eventsLib.log(grant, `perk.${perk.name} revoked`, pgdb)
```
patrickvenetz, line 473:
```suggestion
          eventsLib.log(grant, `perk.${perk.name}.revoked`, pgdb)
```

(use chaining with dots in event names like `role.member.remove`, `email.recipient.recommendations`, etc.)

packages/backend-modules/access/lib/grants.js
```diff
@@ -85,6 +85,35 @@ const grantPerks = async (grant, recipient, campaign, t, pgdb, redis, mail) =>
     { concurrency: 1 },
   )
 
+const revokePerks = async (grant, recipient, campaign, pgdb) =>
```
patrickvenetz, line 88:
```suggestion
const revokePerks = (grant, recipient, campaign, pgdb) =>
```

`async` keywords is not necessary, since a Promise is returned and you're not awaiting anything in _that_ function.

(Can also go for `grantPerks` and lies in my inaptiude to write clean code.)

packages/backend-modules/access/lib/grants.js
```diff
@@ -85,6 +85,35 @@ const grantPerks = async (grant, recipient, campaign, t, pgdb, redis, mail) =>
...
+  Promise.map(
```
patrickvenetz, line 89: To quote grandmother: "Forgive them for they know not what they do!"

I since discovered, instead of `Promise.map(…, { concurrency: 1 })`, there is a readable shorthand: [`Promise.mapSeries(…)`](http://bluebirdjs.com/docs/api/promise.mapseries.html).

Would opt-in to use that shorthand here, and in `grantPerks`, too.

```suggestion
  Promise.mapSeries(
```

(and drop `{ concurreny: 1 }` at end of declaration)

patrickvenetz (changes requested):
I really like this approach.

packages/backend-modules/access/lib/grants.js
```diff
@@ -305,7 +318,7 @@ const request = async (granter, campaignId, payload, t, pgdb, redis, mail) => {
 
   await eventsLib.log(grant, 'request', pgdb)
 
-  const perks = await grantPerks(grant, granter, campaign, t, pgdb, redis, mail)
+  const perks = grantPerks(grant, granter, campaign, t, pgdb, redis, mail)
```
patrickvenetz, line 321:
```suggestion
  const perks = await grantPerks(grant, granter, campaign, t, pgdb, redis, mail)
```

We do need to resolve that Promise before we continue.

packages/backend-modules/access/lib/grants.js
```diff
@@ -432,13 +449,19 @@ const invalidate = async (grant, reason, t, pgdb, mail) => {
         })
       }
 
+      const perks = revokePerks(grant, recipient, campaign, pgdb)
```
patrickvenetz, line 452:
```suggestion
      const perks = await revokePerks(grant, recipient, campaign, pgdb)
```

We do need to resolve this Promise, too, before we can proceed.

patrickvenetz (changes requested):

PR #114, opened 2022-05-17
Add option to script which sends mails to specific segments

fuenkchen:
The send method of mailLog lib already takes an argument, indicating if mails with a specific template name should only be sent once or multiple times to the same email address. 
We want to send invitation mails to cleaned users every three months, no matter if they already received an invitation before or not.

This PR:
* adds another command line arg `onceFor` - a boolean value - which will be used to set the argument in send mail method accordingly


packages/backend-modules/mail/script/sendMailsToSegment/cleanedUserSubscriptionInvitation.js
```diff
@@ -8,6 +8,9 @@ const argv = yargs
   .option('dry-run', {
     default: true,
   })
+  .option('once-for', {
+    default: true,
+  })
```
patrickvenetz, line 13: So, while these changes look good to me, I am pondering if you should push this to a slightly more elaborate version where you could run a script over and over again and it would only send mails if a user has not recevied a resubscribe for the "latested" cleaned record.

Flow in short:

1. User A drops from mailling list, recorded as "cleaned" on 2022-05-16
2. We run script, sending email, using `onceFor.keys.cleanedAt = 2022-05-16` (made up key!)
3. We run script again. Won't be mailing to User A, because record shows still "cleaned" on 2022-05-16 and `onceFor.keys.cleanedAt = 2022-05-16` blocks mailer from sending it again
4. User A resubscribes
5. We run script again, but user is not longer in list.
6. User A drops _again_ from mailling list, recorded as "cleaned" on 2022-07-23
7. We run script again. We will be mailing to User A, because record shows "cleaned" on 2022-07-23 and `onceFor.keys.cleanedAt = 2022-07-23` returns no mailLog record just yet

patrickvenetz (approved):

packages/backend-modules/mail/script/sendMailsToSegment/cleanedUserSubscriptionInvitation.js
fuenkchen: yes that makes totally sense, also to avoid spamming users, who just don't want to resubscribe. Will implement that before merging.

patrickvenetz (approved):

PR #115, opened 2022-05-18
fix Overlay closing when mouse is released over backdrop

trm217:
# Description

On any mouse-down event on the modal, if the mouse was then moved out of the modal and onto the backdrop, on release the modal would close.
This was due to the parent element-registering the click on it's self rather than on it's child.
By moving the onClick-handler, that closes the modal, onto a sibling element of the modal this issue can be circumvented.

# Example
## Before

https://user-images.githubusercontent.com/30313631/169085199-042693a6-9cbe-4fab-9f4b-c08b88fca5a8.mov

## After

https://user-images.githubusercontent.com/30313631/169085709-9a7d7f45-e314-4b16-8d8c-529c188fae29.mov


annatraussnig:
reproduced the issue live and tested the fix on love.

annatraussnig (approved):

PR #116, opened 2022-05-18
Render the App version in the footer

trm217:
# Description

Previously the version number of the app could only be seen in the phone settings or in the app-/play-store.
With this PR the version number will be shown at the very bottom of the footer. (right aligned just as the open source link directly above the version number.

## Screenshot
image

### DIFF
--- a/apps/www/components/Footer/index.js
+++ b/apps/www/components/Footer/index.js
@@ -4,7 +4,6 @@ import compose from 'lodash/flowRight'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import {
-  BrandMark,
   Logo,
   mediaQueries,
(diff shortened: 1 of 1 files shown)


apps/www/components/Footer/index.js
```diff
@@ -337,7 +331,13 @@ const Footer = ({
             <Address t={t} />
           </div>
         </div>
-        <div style={{ textAlign: 'right' }}>
+        <div
+          style={{
+            display: 'flex',
+            flexDirection: 'column',
+            alignItems: 'end',
```
tpreusse, line 338: flex needs to go through `css` to be cross browser compat

ovbm (approved):

PR #118, opened 2022-05-19
Improve UX for selecting credential in discussion-preferences

trm217:
## Description

A onClear option was added to the field component, that renders an "X" on the right. This is now used in the discussion-preferences to allow to clear the credential field with one click.
Additionally the auto-credential functionality was adapted to automatically remove the credential when switching to anonymous. When swichting back from anonymous, the default role is automatically re-set.

### Before

https://user-images.githubusercontent.com/30313631/169315240-4c9b1441-e61c-4649-b468-afac584b30a0.mov

### After

https://user-images.githubusercontent.com/30313631/169315542-21019c86-7fef-471f-833f-ca7c54dc1424.mov


packages/styleguide/src/components/Form/docs.md
```diff
@@ -59,6 +59,20 @@ state: {
...
+```react
+state: {
+  value: 'Eine Rolle'
+}
+---
+<Field
+  label='Rolle'
+  value={state.value}
+  onChange={(_, value) => setState({value})}
+  onClear={() => setState({value:""})} />
```
tpreusse, line 73: do we need a separate function? can we just call `onChange` with an empty string? `onChange` always gets a string anyhow. Also see `browserIconStyle` – needs to be adapted to ensure native icon (like native clear is disabled when we explicitly make a clear button)

packages/styleguide/src/components/Form/Field.tsx
```diff
@@ -68,6 +70,12 @@ const styles = {
     right: 0,
     cursor: 'pointer',
   }),
+  secondaryActionCenter: css({
+    position: 'absolute',
+    top: '50%',
+    right: 0,
+    cursor: 'pointer',
```
tpreusse, line 77: can be removed, is part of `plainButtonRule`

apps/www/components/Discussion/overlays/DiscussionPreferences.js
```diff
@@ -356,6 +357,11 @@ const DiscussionPreferencesEditor = ({
                   setState((curr) => ({
                     ...curr,
                     anonymity: val,
+                    credential:
+                      val === false // if not anonymous set credential
+                        ? userPreference?.credential?.description ??
+                          autoCredential?.description
+                        : '',
```
tpreusse, line 364: Let's keep credential field value when unselecting anonymous:
```suggestion
                      val === false // if not anonymous set credential
                        ? state.credential || userPreference?.credential?.description || autoCredential?.description || ''
                        : '',
```

Also note that the defaulting to an empty string is necessary for the input field to update correctly:
Screenshot 2022-05-25 at 00 05 45
(this happens when setting it to undefined)

tpreusse (commented):
I think the clear icon is great. One thing I'm a bit concerned with it's non obvious that the credential field is cleared when clicking the anon checkbox. If I write a new role `Techie bei einem Magazin` and then set anonymous (which I planned all along) it get's cleared but I might have explicitly written it for my anonymous avatar. IMO the automatic clearing should only happen with auto credentials.

apps/www/components/Discussion/overlays/DiscussionPreferences.js
```diff
@@ -356,6 +357,13 @@ const DiscussionPreferencesEditor = ({
                   setState((curr) => ({
                     ...curr,
                     anonymity: val,
+                    credential:
+                      val === false // if not anonymous keep or set saved credential
+                        ? state.credential ||
+                          userPreference?.credential?.description ||
+                          autoCredential?.description ||
+                          ''
+                        : '',
```
tpreusse, line 366: We should limit the logic here to something really simple that focuses on removing auto credentials when going anonymous, something like this:
```suggestion
                    credential:
                      val && curr.credential === autoCredential?.description 
                        ? '' // remove auto credential when setting anonymous to true
                        :  curr.credential
```
(untested)

Anything more complex probably causes more problems than it solves. It's just not an expected behaviour to mutate other fields when unchecking a box and users might not notice what is being set further above (e.g. on mobile).

tpreusse (changes requested):

apps/www/components/Discussion/overlays/DiscussionPreferences.js
trm217: I adapted your change a bit to unset the credential if it's equal to the initialStates credential.

tpreusse (approved):

PR #122, opened 2022-06-01
Fix billing field length

fuenkchen:
Fixes issue we have with bills and names which are longer than 70 chars.

Therefore we've extended address lib with a char limit check and added validation checks on address form 
Also we made a bit of cleanup and are using address lib by default.

(And another small change sneaked in to have the same char limit on profile biography backend and frontend wise)


packages/backend-modules/republik/lib/address.js
```diff
@@ -1,17 +1,39 @@
...
   if (!address) {
     return
   }
 
+  ensureStringLengthForAddress(
+    address.name,
+    MAX_ADDRESS_NAME_LENGTH,
+    'address/name/tooLong',
+    t,
+  )
```
patrickvenetz, line 19: apps/www/components/Account/AddressForm.js validates `line1` and `city` – but API does not just yet. Since API is responsible for proper data, would validate these inputs before inserting or updating.

patrickvenetz (changes requested):
Admin-Tool in apps/admin/components/Users/Particulars/AddressFieldSet.js has same set of values (in fact, it's build pretty much the same way apps/www.

I find it reasonable to put validators in place there, too. (Even if API would reject length exceeding data.)

We will need a strategy to update existing address rows which violate rules.

fuenkchen:
Test cases:

Profil:
- [ ] Adresse hinzufügen
- [ ] Adresse leeren
- [ ] Adresse ändern

Bestellung:
- [ ] keine Adresse => Rechnungsadresse hinzufügen
- [ ] Adresse da => Rechnungsadresse ändern
- [ ] keine Adresse => Lieferadresse hinzufügen
- [ ] Adresse da => Lieferadresse hinzufügen
- [ ] keine Adresse => Rechnungs- und Lieferadresse hinzufügen
- [ ] Adresse da => Rechnungs- und Lieferadresse ändern (mit Option «Ihre Standardadresse aktualisieren»)
- [ ] Adresse da => Rechnungs- und Lieferadresse ändern (ohne Option «Ihre Standardadresse aktualisieren») 

Admin-Tool:
- [ ] keine Adresse => Adresse hinzufügen
- [ ] Adresse da => Adresse ändern
- [ ] Adresse da => Adresse leeren

PR #123, opened 2022-06-03
fix(Form): checkbox and radio a11y compat

tpreusse:
Solution based on this blog post:
https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons/

Decided to just go with a default outline with automatic color (which works well with dark mode) offset by 3px:

Screenshot 2022-06-03 at 16 41 19

Screenshot 2022-06-03 at 16 41 30

Should also make it easier to select the checkbox in e2e tests ;-)


trm217:
[WAVE browser extension](https://wave.webaim.org/) is happy 👍🏻

trm217:
On `/konto` checkboxes are disabled (until loading is finished) right after they're clicked and the focus is lost , perhaps this might be an issue for screen-readers? 🤔

PR #131, opened 2022-06-12
fix(publikator): save comment id and marker at once

patrickvenetz:
If you pick a marker and then submit an initial comment, picked marker got lost.

https://user-images.githubusercontent.com/331800/173222928-324c5e20-f931-4a16-9432-4f0ad9f8f714.mov

This Pull Request fixes that by storing changes in component state (`parentId`, `marker`) and upon closing overlay, calls `editor.change`.

https://user-images.githubusercontent.com/331800/173222825-ff5dbc47-9688-49a3-81f9-52a52a78c6d4.mov


apps/publikator/components/editor/modules/memo/index.js
```diff
@@ -37,23 +37,18 @@ const Memo = compose(
...
   // noOpen flag added when deserializing
   // only new memos trigger the modal automatically
   useEffect(() => {
     !node.data.get('noOpen') && open()
   }, [])
 
-  const persistData = (key, value) =>
-    editor.change((change) => {
-      change.setNodeByKey(node.key, {
-        data: node.data.set(key, value),
```
annatraussnig, line 50:
```
const persistData = (key, value) =>
    editor.change((change) => {
      change.setNodeByKey(node.key, {
        data: node.data.merge({ [key]: value }),
      })
    })
```

annatraussnig (changes requested):
Thanks for taking this on ❤️ 
I believe just applying `data.merge` instead of `data.set` is enough to solve the problem. No need for states or any of the other changes.

apps/publikator/components/editor/modules/memo/index.js
annatraussnig: I believe this is the only change you need for the bug to go away.

patrickvenetz:
Did give it a spin and it did not work.

Maybe this is a look-me-over-the-shoulder-thing.

annatraussnig (approved):
All our reviewers are currently busy, please hold the line.
All our reviewers are currently busy, please hold the line.
All our reviewers are currently busy, please hold the line.

PR #143, opened 2022-06-28
Bump Next.js to v12.2 and make update middleware to stable API

trm217:
## Description

Update Next.js to v12.2 and update the middleware code to use the stable API.

## TODO

- [ ] Wait for issues with `jose` to be resolved ([#421](https://github.com/panva/jose/issues/421))

### DIFF
--- a/apps/www/next.config.js
+++ b/apps/www/next.config.js
@@ -34,11 +34,6 @@ module.exports = withTM(
     async rewrites() {
       return {
         beforeFiles: [
-          // /front is only accessible via _middleware rewrite
-          {
-            source: '/front',
-            destination: '/404',
-          },
           // _ssr routes are only accessible via rewrites
           {
             source: '/_ssr/:path*',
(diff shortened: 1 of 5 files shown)


apps/www/next.config.js
```diff
@@ -34,11 +34,6 @@ module.exports = withTM(
     async rewrites() {
       return {
         beforeFiles: [
-          // /front is only accessible via _middleware rewrite
-          {
-            source: '/front',
-            destination: '/404',
-          },
```
tpreusse, line 41: why remove this?

trm217: It no longer works in combination with the middleware. 
If we rewrite `/` -> `/front`, with Next.js 12.2 `/` is rewritten to `/404`. 
So in order to fix that I moved rewriting `/front` -> `/404` into the middleware, which doesn't clash with the first mentioned rewrite (which is also done in the middleware)

tpreusse: makes sense, didn't notice it in the middleware.

tpreusse (approved):

PR #146, opened 2022-06-30
Upgrade publikator tooling

trm217:
## Description

TBD

## Tasks

- [x] Replace next-router with native Next.js routing
- [x] Upgrade Next.js
- [x] Configure Jest
- [x] Unify Typescript config comparing to www
- [x] Upgrade to Apollo client V3
- [x] Wait for #143 to be merged
- [ ] React-icons code-splitting
- [x] Fix issues on '/' after fetching data initially


trm217:
@tpreusse All done, but I didn't managed to get react-icons tree-shaking working. It works fine in admin, but not here. no clue why.
Updating the routing was quite messy, since I had to migrate the url from `/repo/{repoId}` to `repo/{owner}/{repo}` (the new router is not happy when we attempt to render a '/' in the path variable. 
We must thoroughly test the all functionalities before we release it.

apps/publikator/components/Auth/SignIn.js
```diff
@@ -238,11 +239,11 @@ class SignIn extends Component {
           </div>
         </form>
         <Label {...styles.hint}>
-          <Link route='legal/privacy'>
+          <Link href='/legal/privacy' passHref>
             <a {...styles.hintA}>{t('signIn/privacy')}</a>
           </Link>
```
tpreusse, line 244: this is a outdated route that never worked in publikator anyhow. let's fix it when changing (maybe also apply in admin tool)

suggestion just as potential solution, needs const import to work:

```suggestion
            <a {...styles.hintA} href={`${FRONTEND_BASE_URL}/datenschutz`}>{t('signIn/privacy')}</a>
```

apps/publikator/components/Auth/SignIn.js
```diff
@@ -238,11 +239,11 @@ class SignIn extends Component {
...
           {' – '}
-          <Link route='faq'>
+          <Link href='/faq' passHref>
             <a {...styles.hintA}>{t('signIn/faq')}</a>
           </Link>
```
tpreusse, line 248: rm faq as not needed in publikator/admin and link does not work

tpreusse (approved):
LGTM. I think the repo id helpers are a good solution. @annatraussnig have a look at `apps/publikator/lib/repoIdHelper.ts` and it's usage just to be aware when merging this back into flyer.

One optional small thing: publikator has had an effect for a while which immediately adds an empty query string. Now this even adds an empty `q` query string. Ideally this could be optimised to only add a query string when not empty:
Screenshot 2022-08-15 at 16 27 17

Issue #72, opened 2022-03-25
Prevent Redirects from opening in the app


ovbm:
Certain redirects, like the google form for goodies, open the app and circumvent the app's whitelist of only opening "republik.ch" domains. Users can get stuck on these pages. 

**Steps to Reproduce**
close the App and open http://republik.ch/s/bestellung from a mail link

**Actual Results**
The app opens with google.com/form....

**Expected Results**
The google form should open in safari

**Possible Solutions**
Handle redirects with a landing page

PR #157, opened 2022-07-28
Redirect Mini Sprint

tpreusse:
We introduce a new `/install` route which roughly follows this logic:
```
- if iOS or Android
	- if inNative App
		- show marketing with probelese/front
		- track visit (incl. utm)
	- else
		- track visit (incl. utm)
		- redirect to App Store on iOS and Play Store on Android
			- open app
				- marketing with probelese
				- track without utms (they are lost)
		- show /probelesen in case page stays open or returning to browser
- else
	- show /probelesen
	- track visit (incl. utm)
```

Additionally following problems were fixed:
- #72 — this was rather tricky since we redirected in three ways to external urls: `next.config.js`, via `getStaticProps` returning a external url to the next redirect engine and in some cases via our `StatusError` component. Newly all external redirects go through our `StatusError` component as `clientRedirection` which also show a loading screen and allow for piwik tracking in our side before people leave. (`/ud/*` routes now come through redirection table again.) 
- pay notes for members, e.g. for `/probelesen` asking member to share, were always hidden via `data-hide-if-active-membership` (css), not they are shown once me is loaded. 
- spacing at article end for member – many br tags create an unnatural whole before member paynotes, margins of format feeds and other elements are totally sufficient – they seem are longer needed

Screenshot 2022-07-28 at 11 30 28

### DIFF
--- a/apps/www/components/Article/Page.js
+++ b/apps/www/components/Article/Page.js
@@ -218,6 +218,7 @@ const ArticlePage = ({
   isPreview,
   markAsReadMutation,
   serverContext,
+  clientRedirection,
 }) => {
   const actionBarRef = useRef()
   const bottomActionBarRef = useRef()
@@ -238,6 +239,7 @@ const ArticlePage = ({
     variables: {
       path: cleanedPath,
     },
+    skip: clientRedirection,
   })
 
   const article = articleData?.article
@@ -440,7 +442,11 @@ const ArticlePage = ({
         render={() => {
           if (!article) {
             return (
-              <StatusError statusCode={404} serverContext={serverContext} />
+              <StatusError
+                statusCode={404}
+                clientRedirection={clientRedirection}
+                serverContext={serverContext}
+              />
             )
           }
           return extract === 'share' ? (
@@ -504,7 +510,11 @@ const ArticlePage = ({
         render={() => {
           if (!article || !schema) {
             return (
-              <StatusError statusCode={404} serverContext={serverContext} />
+              <StatusError
+                statusCode={404}
+                clientRedirection={clientRedirection}
+                serverContext={serverContext}
+              />
             )
           }
 
--- a/apps/www/components/Article/Page.js
+++ b/apps/www/components/Article/Page.js
@@ -775,14 +775,6 @@ const ArticlePage = ({
               {me && hasActiveMembership && (
                 <ArticleRecommendationsFeed path={cleanedPath} />
               )}
-              {(hasActiveMembership || isFormat) && (
-                <>
-                  <br />
-                  <br />
-                  <br />
-                  <br />
-                </>
-              )}
               {!suppressPayNotes && payNoteAfter}
             </>
           )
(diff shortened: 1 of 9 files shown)


apps/www/components/Article/Page.js
```diff
@@ -765,14 +775,6 @@ const ArticlePage = ({
               {me && hasActiveMembership && (
                 <ArticleRecommendationsFeed path={cleanedPath} />
               )}
-              {(hasActiveMembership || isFormat) && (
```
ovbm, line 768: eliminating <br>. I'm IN!

apps/www/pages/install.js
```diff
@@ -0,0 +1,62 @@
...
+
+const InstallNativeAppPage = ({ clientRedirection }) => (
+  <Frame raw>
+    <StatusError status={404} clientRedirection={clientRedirection} />
+  </Frame>
+)
+
+export default InstallNativeAppPage
+
+const getQueryString = (req) => {
```
ovbm, line 19: why not use `import { ParsedUrlQuery } from 'querystring'` or URLSearchParams API ?

ovbm (approved):

apps/www/components/StatusError/index.js
```diff
@@ -14,66 +14,155 @@ import Meta from '../Frame/Meta'
...
+    return url?.replace(/^https?:\/\/www\./, '')
+  }
+}
+const appendQueryString = (target, queryString) => {
+  return `${target}${
+    queryString ? `${target.includes('?') ? '&' : '?'}${queryString}` : ''
+  }`
+}
+
+const StatusError = ({
```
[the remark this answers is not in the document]
tpreusse: It probably never was the right name – it always tried to find a redirects on 404s and did them – showing a loader without a message client side. Ideas? `StatusErrorAndRedirects`?

apps/www/pages/install.js
tpreusse: Would have to use something like:
```
import {format,parse} from url
format({
  pathname: '/',
  query: parse(req.url).query
})
format({
  pathname: '/probelesen',
  query: parse(req.url).query
})
```

But no real reason, just felt like it was easy and predictable enough to do with simple string operations.

apps/www/components/StatusError/index.js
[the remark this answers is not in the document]
ovbm: Maybe RedirectStatus, and Error ist just one possible status?

apps/www/pages/install.js
ovbm: got it

apps/www/components/StatusError/index.js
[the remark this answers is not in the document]
tpreusse: It also handles 503 and in most cases – everywhere except `install.js` – first a 404 / document not found needs to happen before it does redirect. Maybe `HTTPStatus` or `PageRetrievalStatus` would be good but I think I'll keep that in mind for the future and leave it for now.

PR #159, opened 2022-07-29
build(nextjs-apollo-client): drop rollup

tpreusse:
similar setup to styleguide

tested on IE11 to work

(became necessary after rollup [blocked my deploy](https://piratinnen.slack.com/archives/C8SE5KMSN/p1659119616012969?thread_ts=1659117548.181169&cid=C8SE5KMSN), was unable to debug and identify the issue after multiple hours)

PR #160, opened 2022-08-02
Fix rollup.js issues in nextjs-apollo-config

trm217:
## Description

Reverts #159 which removed the rollup setup from `nextjs-apollo-config`.
It seems the problems where caused by the latest patch release of `@rollup/typescript-plugin` (v8.3.4) which was released at the end of last week (the following [changes](https://github.com/rollup/plugins/pull/1201) seemed to cause the issue). After locking the version of the rollup-plugin the issue we saw when deploying to staging, stopped.

Besides that the TS-config now the defines the source folder explicitly & the rollup-plugin order was changed to run the commonjs-plugin before the typescript-plugin (which should've been the case all along).


trm217:
TS seems to work fine in editors after moving to compiling only with `tsc`.

PR #163, opened 2022-08-03
Flyer editor

annatraussnig:
Publikator Todos (higher prio towards the top):

- [x] setup to share components using graphql between styleguide and publikator
- [x] staging environment
- [x] review apps/publikator/components/Edit/index.js 
  - [x] refactor to just use repo route for both schemas
  - [ ] refactor local storage and uncommited changes pings into a custom hook
- [x] phase overview / modal for workflow
- [x] integrate new metadata form
- [x] preview page using [SlateRender](http://localhost:3003/editor/flyer#rendering)
- [x] preview for publish page
- [x] edit page details: style warnings, handle archived banner, readonly mode, add info when you are x versions behinds
- [x] slate: edit buttons
- [x] article teaser component
- [ ] error boundary configuration: reload based on localstorage
- [x] diff view
- [x] calendar integration
- [x] editor forms:
  - [x] use correct path in repo search so that backend can replace it
  - [x] add repo search functionality to link box


apps/publikator/components/Edit/index.js
```diff
@@ -0,0 +1,522 @@
...
+  useEffect(() => {
+    if (prevUncommittedChanges?.users !== uncommittedChanges?.users) {
+      updateActiveUsers()
+    }
+  }, [uncommittedChanges])
+
+  // new route, new store
+  useEffect(() => {
+    initStore()
+  }, [commitId, repoId])
```
tpreusse, line 129: store and uncommitted changes could we put into a custom hook

packages/styleguide/src/components/Editor/config/elements/link.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  // Directly updating the link props causes issues
+  // with the normalizer and the cursor.
+  useEffect(() => {
+    return () => {
+      Transforms.insertText(editor, textRef.current, {
+        at: path,
+      })
+    }
+  }, [])
```
tpreusse, line 39: This transform breaks any marks used inside the link text, e.g. CO<sub>2</sub> will not survive the link modal unmounting.

I would recommend to remove the feature of editing text inside of the modal. Editing within slate is sufficient IMO.

packages/styleguide/src/components/Editor/config/elements/articlePreview/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+              if (meta.title) {
+                Transforms.insertText(editor, meta.title, {
+                  at: path.concat([1, 0]),
+                })
+              }
+              if (meta.shortTitle) {
+                Transforms.insertText(editor, meta.shortTitle, {
+                  at: path.concat([1, 1]),
+                })
+              }
```
tpreusse, line 68: are you sure you want to use the `shortTitle` here? it would be more logical to be `description` – `title` & `description` are made to work together. `shortTitle` is made to work standalone.

annatraussnig: You are right, I changed it.

packages/styleguide/src/components/Editor/config/elements/link.tsx
annatraussnig: I find it quite useful to be able to see and edit everything in one place. I went for a compromise, for now:

- only apply the transform if the text has been changed
- warning about formatting in the form

PR #166, opened 2022-08-04
129 a11y communicate activeselected state of navigationselection elements

trm217:
## Description

TBD


packages/styleguide/src/components/Tabs/Scroller.tsx
```diff
@@ -1,5 +1,5 @@
 import { css } from 'glamor'
-import React, { useRef, useState, useEffect } from 'react'
+import React, { useRef, useState, useEffect, Children } from 'react'
```
ovbm, line 2: Changes in the Scroller don't seem related to acessibility. I don't have time to check these now. Has this been tested? Seems to me that it would change layout behavior.

trm217: You're right, that wasn't an issue mentioned in the issue. However this changes were made to make the tab buttons to be easier accessbile by wrapping them in a list as well as passing certain a11y-attributes. It shouldn't change the functionality, but if you could test it as well that'd be cool.

PR #169, opened 2022-08-05
136 modals sollten barrierefrei umgesetzt sein

trm217:
TBD


trm217:
## dev-note

Adding a focus-trap to our ArticleGallery component isn't working due to no focusable images being available in the initial render.
In order for focus-trap to work we must adapt the code to only render once the an image has already been loaded.

trm217:
@tpreusse Once you have time, I would like to discuss the focus-styling with you.
We have a defined styling for the focus-state on a bunch of components, however for example on Link components we don't.
I'd be in favor of using a default focus styling such as: `{ outline: '2px dotted black', outlineOffset: '4px' }`. However for ui such as Links I believe a individual styling would be nicer (such as making the Link bold on focus).

PR #175, opened 2022-08-11
Bump Next.js to v12.2.4 and refactor middleware me.roles fetching

trm217:
## Description

Bump Next.js to v12.2.4 and refactor the API-Call, that is made to the api to retrieve me.roles + the user-token, to be a GET-request.

### DIFF
--- a/apps/admin/package.json
+++ b/apps/admin/package.json
@@ -24,7 +24,7 @@
   ],
   "dependencies": {
     "@apollo/client": "~3.5.10",
-    "@next/bundle-analyzer": "^12.2.3",
+    "@next/bundle-analyzer": "^12.2.4",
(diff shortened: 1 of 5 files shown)


trm217:
Doesn't work after all, seems the mentioned fix doesn't actually get rid of this problem.
When trying to access '/' while only the connect.sid cookie is present, the middleware crashes with the above mentioned error. However after the user sees the 500 error-page, the user-token is loaded, which makes a subsequent reload, deliver the correct page.

Let's keep track on the [problem in the Next.js mono-repo](https://github.com/vercel/next.js/issues/38963). In case it doesn't, it might make more sense to look into a gateway page, that is delivered incase the user-token isn't available when the middleware is being hit.

PR #177, opened 2022-08-16
feat(www): add citation meta tags for zotero

trm217:
## Description

Adds the following meta-tags, based on the [Zotero documentation](https://www.zotero.org/support/dev/exposing_metadata):

- citation_title
- citation_journal_title
- citation_date
- citation_authors

### DIFF
--- a/apps/www/components/Article/metadata.js
+++ b/apps/www/components/Article/metadata.js
@@ -72,6 +72,26 @@ const getJSONLDs = (meta) => {
   }
 }
 
+function getCitationMetaData(meta) {
+  const authors = meta.contributors
+    .filter((contributor) => contributor?.kind?.includes('Text'))
+    .map((contributor) => {
+      const parts = contributor.name.split(' ')
+      // Convert the name into the following format `Surname, Given Name(s)`
+      return `${parts[parts.length - 1]}, ${parts
+        .slice(0, parts.length - 1)
+        .join(' ')}`
+    })
+    .join('; ')
+
+  return {
+    citation_title: meta.title,
+    citation_journal_title: PUBLIC_BASE_URL,
+    citation_date: meta.publishDate,
+    citation_authors: authors,
+  }
+}
+
 export const getMetaData = (documentId, meta) => {
   const shareImage =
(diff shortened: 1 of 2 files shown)


apps/www/components/Article/metadata.js
```diff
@@ -72,6 +72,26 @@ const getJSONLDs = (meta) => {
...
 
+function getCitationMetaData(meta) {
+  const authors = meta.contributors
+    .filter((contributor) => contributor?.kind?.includes('Text'))
+    .map((contributor) => {
+      const parts = contributor.name.split(' ')
+      // Convert the name into the following format `Surname, Given Name(s)`
+      return `${parts[parts.length - 1]}, ${parts
+        .slice(0, parts.length - 1)
+        .join(' ')}`
```
tpreusse, line 83:
```suggestion
      return contributor.name
```

Since some people have multiple word names etc. I'd just give it as is.

There are other examples without the comma split:
https://www.w3.org/WAI/RD/wiki/Indexing

apps/www/components/Article/metadata.js
```diff
@@ -72,6 +72,26 @@ const getJSONLDs = (meta) => {
...
+    })
+    .join('; ')
+
+  return {
+    citation_title: meta.title,
+    citation_journal_title: PUBLIC_BASE_URL,
```
tpreusse, line 89: What's the thinking behind putting in the url here instead of `Republik Magazin` or something like this?

apps/www/components/Article/metadata.js
```diff
@@ -72,6 +72,22 @@ const getJSONLDs = (meta) => {
...
+  const authors = meta.contributors
+    .filter((contributor) => contributor?.kind?.includes('Text'))
+    .map((contributor) => {
+      return contributor.name
+    })
+    .join('; ')
+
+  return {
+    citation_title: meta.title,
+    citation_journal_title: PUBLIC_BASE_URL,
```
trm217, line 85:
```suggestion
    citation_journal_title: 'Republik',
```

PR #181, opened 2022-08-18
chore: bump node version requirements to v16

trm217:
## Description
TBD


trm217:
Seems like turbo is no longer blocking us from upgrading from Node.js v14 to v16.
I suppose we'd have to test each app on heroku, if it all still works as expceted.
@patrickvenetz could you look into potential impacts for the api?

trm217:
All next.js app were able to build and deploy successfully to their staging-environment using Node v16.17

patrickvenetz (approved):
LGTM

Tested api and assets server on staging.

PR #182, opened 2022-08-18
Fixes n Stitches

tpreusse:
Fixes three major bugs
- ensure tracking of initial page views on SSG pages
- fix shouldIgnoreClick import in header and paynote
- fix report error formatting


apps/www/components/Track.js
```diff
@@ -88,6 +95,7 @@ const Track = () => {
       lastMeRef.current = me
     }
     payload.disable(me?.activeMembership)
+    throw new Error('test')
```
trm217, line 98: 👀

apps/www/components/Track.js
```diff
@@ -19,7 +19,14 @@ const trackRoles = (me) =>
     me ? [].concat(me.roles).sort().join(' ') || 'none' : 'guest',
   ])
 
+let lastTrackedUrl
```
trm217, line 22: Any particular reason this code is outside of the component?

apps/www/components/Track.js
trm217, line 98: 👀

trm217 (changes requested):
Looks good to me besides the error being thrown

apps/www/components/Track.js
tpreusse: sorry commited the test by accident 😱

apps/www/components/Track.js
tpreusse: don't want to use a react ref and I prefer placing functions outside of react components whenever possible to avoid shadowing issues

PR #189, opened 2022-08-30
feat(readAloud): landing Page

ovbm:
Screenshot 2022-08-30 at 15 21 41

Todo: 
- [x] Copy
- [x] Test if removal of buttonWidth didn't cause mayhem


ovbm:
Tested ButtonWidth: Removal of fixed size doesn't change layout if no custom buttonLabel is set.

apps/www/components/Auth/EmailForm.js
```diff
@@ -25,7 +25,6 @@ const styles = {
     flexGrow: 1,
   }),
   button: css({
-    width: 160,
```
ovbm, line 28: Fyi: tested and doesn't change layout if no custom buttonLabel is provided. Removing it allows us to add longer button labels to the Email from: eg. "Benachrichtigen"

trm217: Are you sure there are no side-effects? If we just change it to minWidth we can ensure that the behaviour remains the same for short text but still allow for longer text.

trm217 (changes requested):
See The "you're subscribed" label seems to be broken in all browsers (screenshot below). I guess a 'display: inline-block' must be missing in the element wrapping the checkbox + label

image

apps/www/components/Auth/EmailForm.js
ovbm: I am pretty sure. minWdith would also prevent shorter labels to look consistent

patrickvenetz:
(The Vollständigkeit halber:) `inline` due to Editorial leads which are preceeded by a subject line

Bildschirmfoto 2022-09-02 um 14 16 05

trm217:
Once again our "component library" is in no way predictable…

apps/www/components/Auth/EmailForm.js
trm217: Well until now the width was set to 160 fixed, so for labels that take up less space, the buttons would look different with this change.

trm217 (approved):

PR #190, opened 2022-08-30
Newsletter signup, CLIMATE, READALOUD

patrickvenetz:
This Pull Request adds two now Mailchimp segments to code base: `CLIMATE` (Klimalabor-Newsletter) and `READALOUD` («Hört, Hört» notification once released).

Both segments are invisible and won't be listed. API however returns subscription if queried by name:

```gql
{
  user(slug: "someone") {
    newsletterSettings {
      subscriptions(name: CLIMATE) {
        name
        subscribed
      }
    }
  }
}
``` 

This enables `Auth/NewsletterSubscriptions` to handle invisible Mailchimp segments if a name is provided (used when i. e. `Document.meta.newsletter.name` is set).

It then exposes `Auth/NewsletterSignUp` as a "Dynamic Component".

And it also refactors `Auth/Notifications` and `Auth/MacNewsletterSubscription` to allow for Mailchimp segment specific title and short text:

Bildschirmfoto 2022-08-30 um 15 31 54

### DIFF
--- a/apps/www/components/Account/NewsletterSubscriptions/index.js
+++ b/apps/www/components/Account/NewsletterSubscriptions/index.js
@@ -35,7 +35,7 @@ export const UPDATE_NEWSLETTER_SUBSCRIPTION = gql`
 `
 
 export const NEWSLETTER_SETTINGS = gql`
-  query myNewsletterSettings {
+  query myNewsletterSettings($newsletterName: NewsletterName) {
     me {
       id
       newsletterSettings {
@@ -47,7 +47,7 @@ export const NEWSLETTER_SETTINGS = gql`
 `
 
 const NewsletterSubscriptions = ({ t, isMember, free, onlyName }) => (
-  <Query query={NEWSLETTER_SETTINGS}>
+  <Query query={NEWSLETTER_SETTINGS} variables={{ newsletterName: onlyName }}>
     {({ loading, error, data }) => {
       if (loading || error) {
         return <Loader loading={loading} error={error} />
--- a/apps/www/components/Account/NewsletterSubscriptions/index.js
+++ b/apps/www/components/Account/NewsletterSubscriptions/index.js
@@ -35,7 +35,7 @@ export const UPDATE_NEWSLETTER_SUBSCRIPTION = gql`
 `
 
 export const NEWSLETTER_SETTINGS = gql`
-  query myNewsletterSettings($newsletterName: NewsletterName) {
+  query myNewsletterSettings($onlyName: NewsletterName) {
     me {
       id
       newsletterSettings {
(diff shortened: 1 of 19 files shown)


apps/www/components/Account/NewsletterSubscriptions/index.js
```diff
@@ -47,7 +47,7 @@ export const NEWSLETTER_SETTINGS = gql`
 `
 
 const NewsletterSubscriptions = ({ t, isMember, free, onlyName }) => (
-  <Query query={NEWSLETTER_SETTINGS}>
+  <Query query={NEWSLETTER_SETTINGS} variables={{ newsletterName: onlyName }}>
```
annatraussnig, line 50: i know this is from before, but I find `onlyName` to be a confusing, erm, name. Maybe it could be called `newsletterName`everwhere?

packages/backend-modules/mail/NewsletterSubscription.js
```diff
@@ -1,9 +1,8 @@
 const createNewsletterSubscription = (interestConfigurationMap) => ({
-  buildSubscription(userId, interestId, subscribed) {
-    const interestConfiguration = this.interestConfiguration(interestId)
-    const name = interestConfiguration.name
+  buildSubscription(userId, interestId, subscribed, roles) {
+    const { name, ...rest } = this.interestConfiguration(interestId)
     const id = Buffer.from(userId + name).toString('base64')
-    return { name, id, subscribed }
+    return { userId, subscribed, roles, name, ...rest, id }
```
annatraussnig, line 5: sylistically speaking, I'd put `...rest` at the end of the enumeration.

packages/backend-modules/republik/script/generateNewsletterHmac.ts
```diff
@@ -0,0 +1,36 @@
...
+const confirmLink = `${FRONTEND_BASE_URL}/mitteilung?type=newsletter&name=${name}&subscribed=1&email=${base64u.encode(
+  email,
+)}&mac=${hmac}&context=newsletter`
+
+console.log({
+  email,
+  name,
+  hmac,
+  confirmLink,
+})
```
annatraussnig, line 36: is this meant to be? (In FE random console.logs aren't exactly welcome)

annatraussnig (approved):
Frontend looks good to me. Backend looks fantastic, but who am I to say...?

packages/backend-modules/republik/script/generateNewsletterHmac.ts
patrickvenetz: Yes, this is a CLI script and prints results via `console.log`.

packages/backend-modules/mail/NewsletterSubscription.js
patrickvenetz: I usually agree. However, in this instance, I do want anything in `...rest` to be overwritten by other variables. (Did put `...rest` at begining now.)

apps/www/components/Account/NewsletterSubscriptions/index.js
patrickvenetz: I think what supports the confusen is the switch to another wording once more (`newsletterName`) – I reworked it to keep `onlyName` from component prop down to query.

PR #203, opened 2022-09-10
fix grammar in README section "Postgresql in docker"

das-g:


patrickvenetz (approved):
LGTM, and thank you @das-g

PR #210, opened 2022-09-14
feat: refetch flyer path every day after 5am

tpreusse:
Caveat: currently has no clean up. Feels unnecessary in prd but in dev it could go crazy – with hot reload multiple timeouts could be running. A clean up option I could imagine are ref counting:

```
let refs = 0
let timeoutId
export const useFlyerMeta = () => {
  useEffect(() => {
    refs += 1
    if (timeoutId === undefined) timeoutId = setTimeout()
    // ...
    return () => {
      refs -= 1
      if (refs === 0) clearTimeout(timeoutId)
    }
  }, [])
}
```

### DIFF
--- a/apps/www/lib/apollo/miniNavi.js
+++ b/apps/www/lib/apollo/miniNavi.js
@@ -1,4 +1,5 @@
 import { gql, useQuery } from '@apollo/client'
+import { useEffect } from 'react'
 
 export const miniNaviQuery = gql`
   query miniNavi {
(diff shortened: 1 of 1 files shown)


apps/www/lib/apollo/miniNavi.js
```diff
@@ -15,8 +16,44 @@ export const miniNaviQuery = gql`
...
+      PUBLISH_HOURS,
+      PUBLISH_MINUTES,
+    )
+    const nextUpdateTomorrow = new Date(
+      now.getFullYear(),
+      now.getMonth(),
+      now.getDate() + 1,
+      PUBLISH_HOURS,
+      PUBLISH_MINUTES,
+    )
```
trm217, line 45: I have doubts that this work as expected. Just adding +1 to the day would yield invalid dates at the end of every month.

I would compute the next date simply by adding 1 day in ms to the current timestamp.

(see [this fiddle](https://jsfiddle.net/78v2y9pa/1/) to see the issue)

```suggestion
    const DAY_IN_MS = 24 * 60 * 60 * 1000
    const nextUpdateTomorrow = new Date(updateTimeToday.getTime() + DAY_IN_MS)
```

apps/www/lib/apollo/miniNavi.js
```diff
@@ -15,8 +16,44 @@ export const miniNaviQuery = gql`
...
+    const nextUpdateMs =
+      now < updateTimeToday ? updateTimeToday - now : nextUpdateTomorrow - now
+    const refresh = () => {
+      refetch()
+      // recursion 24h later again
+      refetchTimeoutId = setTimeout(() => {
+        refresh()
+      }, 24 * 60 * 60 * 1000)
```
trm217, line 53:
```suggestion
      }, DAY_IN_MS)
```

trm217: Only apply if the above suggestion is applied too

trm217 (dismissed):
Computing of updateTomorrow has to be changed, but aside from that LGTM

trm217 (approved):
Code looks good to me 👍🏻 
(couldn't test it locally though, I assume it worked for you based on the last commit)

PR #213, opened 2022-09-15
Pass identifiers to the apollo-client across our web-clients

trm217:
- feat(nextjs-apollo-config): allow passing a name and version to the apollo-client
- feat(www): pass buildId to public-runtime config and use it as client-version
- feat(clients): pass buildId to publicRuntimeConfig to pass to apollo-client

### DIFF
--- a/apps/admin/lib/apollo/index.ts
+++ b/apps/admin/lib/apollo/index.ts
@@ -1,12 +1,15 @@
+import getConfig from 'next/config'
 import {
   createApolloClientUtilities,
   makeWithDefaultSSR,
 } from '@republik/nextjs-apollo-client'
 import { API_URL } from '../../server/constants'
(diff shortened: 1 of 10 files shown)


tpreusse:
Very cool. Could we also add this to our report error? https://github.com/republik/plattform/blob/cf77781eacc2283bc68cd5754db5a769172a855b/apps/www/lib/errors.js#L17

Maybe after the current location like this:
```js
  fetch('/api/reportError', {
    method: 'POST',
    body: `${context}\n${window.location.href} (build ${buildId})\n${error}`,
  })
```

trm217:
Yes, that should work. I will add it in a bit 👍🏻

patrickvenetz (approved):
LGTM

PR #214, opened 2022-09-16
fix(Feed): don't render actionbar on newsletter and flyer

ovbm:
Screenshot 2022-09-16 at 10 26 58

### DIFF
--- a/apps/www/components/ActionBar/index.js
+++ b/apps/www/components/ActionBar/index.js
@@ -139,7 +139,10 @@ const ActionBar = ({
 
   const isSeriesOverview = meta && meta.series?.overview?.id === document?.id
   const hasPdf = meta && meta.template === 'article' && !isSeriesOverview
-  const notBookmarkable = meta && meta.template === 'page'
+  const notBookmarkable =
+    meta?.template === 'page' ||
+    meta?.template === 'flyer' ||
+    meta?.template === 'editorialNewsletter'
   const isDiscussion = meta && meta.template === 'discussion'
   const emailSubject = t('article/share/emailSubject', {
     title: document.meta.title,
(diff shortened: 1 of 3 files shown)


trm217 (approved):

patrickvenetz:
To be consistent, [implement check in search Document results](https://github.com/republik/plattform/blob/d98376353f76724a78b955295fbe3f3bcec4768f/apps/www/components/Search/DocumentResult.js#L88) – a `TeaserFeed` with `bar` prop –, too.

But after digging, it seems more prudent to implement that logic in `ActionBar` itself: It contains most business logic on whether to render an element or not, [like bookmark element should not be rendered if `meta.template === "page"`](https://github.com/republik/plattform/blob/d98376353f76724a78b955295fbe3f3bcec4768f/apps/www/components/ActionBar/index.js#L325).

patrickvenetz (changes requested):

ovbm:
thx for review. Logic is now in action bar. Actionbar also has the option now to not render anything (also no container) if no Items are available.

apps/www/components/ActionBar/index.js
```diff
@@ -242,7 +245,7 @@ const ActionBar = ({
         />
       ),
       modes: ['articleOverlay', 'feed', 'bookmark', 'seriesEpisode'],
-      show: !!document,
+      show: !!document && document.userProgress,
```
ovbm, line 248: this was a bug that I only detected now. Actionbar would render an empty UserProgress component, even if userProgress was null

patrickvenetz: That was because [UserProgress](https://github.com/republik/plattform/blob/f1d5bf64bd7de69b57b2f6dd800cc12a1604fac0/apps/www/components/ActionBar/UserProgress.js#L78-L82) will return `null` if `userProgress` is falsy and thus attempting to [render an `IconButton` with too few props](https://github.com/republik/plattform/blob/f1d5bf64bd7de69b57b2f6dd800cc12a1604fac0/apps/www/components/ActionBar/index.js#L492-L496)?

patrickvenetz (approved):
LGTM

PR #216, opened 2022-09-16
fix(styleguide/templates): don't render series nav within the article if series-meta is null

trm217:


trm217:
I will create an issue describing the odd Apollo behaviour in a bit.

patrickvenetz (approved):
LGTM

PR #217, opened 2022-09-18
refactor(discussions): replace assembleTree fn w/ faster version

patrickvenetz:
This Pull Request replaces `assembleTree`, which builds a discussion comments tree, with a faster, more common pattern.

### DIFF
--- a/packages/backend-modules/discussions/graphql/resolvers/Discussion/comments.js
+++ b/packages/backend-modules/discussions/graphql/resolvers/Discussion/comments.js
@@ -6,50 +6,80 @@ const getSortKey = require('../../../lib/sortKey')
 
 const THRESHOLD_OLD_DISCUSSION_IN_MS = 1000 * 60 * 60 * 72 // 72 hours
 
-const assembleTree = (_comment, _comments) => {
-  const coveredComments = []
+const assembleTree = (
+  comments,
+  { parentId = false, includeParent = false },
+) => {
+  const rootNodes = []
 
-  const _assembleTree = (comment, comments, depth = -1) => {
-    const parentId = comment.id || null
-    comment._depth = depth
-    comment.comments = {
-      nodes: _.remove(
-        comments,
-        (c) =>
-          (!parentId && !c.parentIds) ||
-          (c.parentIds && c.parentIds.slice(-1).pop() === parentId),
-      ),
-    }
-    comment.comments.nodes = comment.comments.nodes
-      .map((c) => {
-        coveredComments.push(c)
-        return c
-      })
-      .map((c) => _assembleTree(c, comments, depth + 1))
-    return comment
+  const ref = comments.reduce((acc, comment, i) => {
+    acc[comment.id] = i
+    return acc
+  }, {})
+
+  comments
+    .filter((comment) => {
+      const missingParentId = !parentId
+      const isCommentParent = comment.id === parentId
+      const isCommentParentChid = comment.parentIds?.includes(parentId)
+
+      return missingParentId || isCommentParent || isCommentParentChid
+    })
+    .forEach((comment) => {
+      const isCommentRoot = !comment.parentIds?.length
+      const isParentCommentAsRoot = comment.id === parentId
+
+      if (isCommentRoot || isParentCommentAsRoot) {
+        rootNodes.push(comment)
+        return
+      }
+
+      const parentCommentId = comment.parentIds[comment.parentIds.length - 1]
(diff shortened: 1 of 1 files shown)


patrickvenetz:
🐛 `orderBy: REPLIES` not working as expected

packages/backend-modules/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -6,50 +6,80 @@ const getSortKey = require('../../../lib/sortKey')
...
+  const ref = comments.reduce((acc, comment, i) => {
+    acc[comment.id] = i
+    return acc
+  }, {})
+
+  comments
+    .filter((comment) => {
+      const missingParentId = !parentId
+      const isCommentParent = comment.id === parentId
+      const isCommentParentChid = comment.parentIds?.includes(parentId)
```
fuenkchen, line 24:
```suggestion
      const isCommentParentChild = comment.parentIds?.includes(parentId)
```

packages/backend-modules/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -6,50 +6,80 @@ const getSortKey = require('../../../lib/sortKey')
...
+
+      return missingParentId || isCommentParent || isCommentParentChid
```
fuenkchen, line 26:
```suggestion
      return missingParentId || isCommentParent || isCommentParentChild
```

fuenkchen (commented):
In general it lgtm. One suggestion, because I always think it's a bit difficult to parse what `assembleTree` is doing in detail: add a short explanatory comment what's happening by filtering and looping over comments. Maybe another comment for `buildConnections` would be helpful as well.

hdahlheim:
@lucianak, we should discuss again whether having so much code as part of the resolver files is what we want moving forward, or if we should move part of it back into lib and convert it to TypeScript.

packages/backend-modules/discussions/graphql/resolvers/Comment.js
```diff
@@ -227,67 +227,51 @@ module.exports = {
...
-      } else {
-        anonymous = false
+    const { discussion, commenterPreferences } = await Promise.props({
+      discussion: loaders.Discussion.byId.load(discussionId),
+      commenterPreferences:
+        loaders.Discussion.Commenter.discussionPreferences.load({
+          userId,
+          discussionId,
+        }),
+    })
```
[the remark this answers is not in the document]
jstcki: Does Promise.props do the same thing as
```
const [discussion, commenterPreferences] = await Promise.all([..., ...])
```
??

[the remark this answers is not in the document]
lucianak: As far as I understand the [documentation](http://bluebirdjs.com/docs/api/promise.props.html) it does: 

> Like [.all](http://bluebirdjs.com/docs/api/all.html) but for object properties or Maps* entries instead of iterated values. Returns a promise that is fulfilled when all the properties of the object or the Map's' values** are fulfilled.

But I agree with Henning that it's probably not worth to add the bluebird dependency for this.

packages/backend-modules/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -2,54 +2,98 @@ const graphqlFields = require('graphql-fields')
...
+  return {
+    comments: {
+      nodes: rootNodes,
+    },
+  }
 }
 
-const measureTree = (comment) => {
+const buildConnections = (comment) => {
   const { comments } = comment
```
lucianak, line 76: Maybe the naming could be improved here for more clarity

packages/backend-modules/discussions/graphql/resolvers/Discussion/comments.js
```diff
@@ -259,30 +333,32 @@ module.exports = async (discussion, args, context, info) => {
...
 
-  const discussionTotalCount = comments.length
+  // fetch comment stubs
+  const partials = await pgdb.query(
+    [
+      `SELECT ${fields.join(', ')}`,
+      'FROM comments c',
+      tag && 'LEFT JOIN comments cr ON cr.id = (c."parentIds"->>0)::uuid',
+      'WHERE c."discussionId" = :discussionId',
+      tag && `AND (c.tags @> '"${tag}"' OR cr.tags @> '"${tag}"')`,
```
lucianak, line 354: tag should also be included as query parameter if possible

PR #221, opened 2022-09-20
More Journal Fixes

annatraussnig:
- Journal title when created through calendar is now based on publish date. (Before all journals were named `Journal` and that's not very enlightening.)
Screen Shot 2022-09-20 at 17 32 29

- progagate readonly props to the editor itself, hiding the toolbar when in readonly mode.
Screen Shot 2022-09-21 at 10 39 26

- adjust end salutation in template

- remove trennstrich when not rendering bgrüssung (now also shown as such in preview)
Screen Shot 2022-09-21 at 15 39 34
ring begrüssung (request from Oli)


patrickvenetz (approved):
LGTM

patrickvenetz:
Not part of this PR, but locking mechanism in flyer editor is not working as it does with old editor. Probably something we've to spec.

PR #222, opened 2022-09-21
Reintroduce flyer path

tpreusse:
Our `NavLink`s to the flyer are rendered behind a `isMember` or  `hasOverviewNav` defined as `hasAccess && wantOverviewNav` check. But I ended up using a hook to query the current flyer path and used it atop of the `SecondaryNav` component which is rendered for everyone.

To remedy this I've introduced two measure:
1. move the hook to a new flyer nav link component that is only rendered when really needed
2. skip the query in the hook when `!hasAccess` and throw an error in dev that the hook should only be used in a sub tree that has access

Edit: also contains unrelated `SeriesNav` changes to fix react warnings I've noticed in my console while testing.

### DIFF
--- a/apps/www/components/Frame/Popover/FlyerNavLink.js
+++ b/apps/www/components/Frame/Popover/FlyerNavLink.js
@@ -1,0 +1,14 @@
+import NavLink from './NavLink'
+import useFlyerMeta from './useFlyerMeta'
+
+const FlyerNavLink = ({ children, ...props }) => {
+  const flyerMeta = useFlyerMeta()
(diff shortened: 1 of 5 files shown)


trm217 (approved):
Looks good to me, the only think I am questioning, is that the formatColor prop was previously only passed in SecondaryNav.js.
Now with this change is passed in Nav.js as well. 
(This causes the hover-behaviour as seen below)
image

@tpreusse Was that change intended?

tpreusse:
@trm217 thanks for the review. No the change was not intentional – thanks for reporting. I will limit the `formatColor` to the mini nav.

PR #233, opened 2022-10-06
Disable Prefetch in Header

tpreusse:
`Frame/Header.js` is currently responsible that every not signed in article load triggers a `/angebote` prefetch including hitting the backend.

Screenshot 2022-10-06 at 23 38 14

## Afterwards

Screenshot 2022-10-06 at 23 44 42

## Misc

Also disable section and cancel link prefetch—those pages are rarely used or should be slow 😇 
Buy links in expanded navis still prefetch but only after clicking the navi. Pay notes do not use next link and have never prefetched.

### DIFF
--- a/apps/www/components/Frame/Header.js
+++ b/apps/www/components/Frame/Header.js
@@ -312,7 +312,7 @@ const Header = ({
                   </a>
                 </Link>
               ) : (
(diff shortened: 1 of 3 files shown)


trm217 (approved):
Merci Thomas

PR #234, opened 2022-10-07
Hoert hoert frontend launch (19th of Oct)

ovbm:
https://www.notion.so/republikmagazin/Front-End-H-rt-H-rt-0af7fcfa8f47430bbc6d196fa67f9cc2


apps/www/components/Audio/AudioPlayer/AudioPlayer.tsx
```diff
@@ -16,6 +16,7 @@ import { AnimatePresence, motion } from 'framer-motion'
 import { css } from 'glamor'
 import AudioPlaybackElement from './AudioPlaybackElement'
 import { useUserAgent } from '../../../lib/context/UserAgentContext'
+import { useAudioContext } from '../AudioProvider'
```
ovbm, line 19: not used?

annatraussnig: good catch

apps/www/components/Frame/Header.js
```diff
@@ -8,8 +8,8 @@ import {
   HeaderHeightProvider,
   useColorContext,
   shouldIgnoreClick,
+  BackIcon,
 } from '@project-r/styleguide'
```
ovbm, line 12: It was a bit hard to understand that the Toggle now is the AudioPlayer-Toggle. Maybe rename?

apps/www/components/Frame/SecondaryNav.js
```diff
@@ -140,7 +152,7 @@ const styles = {
       '::after': {
         ...fontStyles.sansSerifMedium,
         display: 'block',
-        content: 'attr(title)',
+        content: 'attr(data-placeholder)',
```
ovbm, line 155: clever :D

apps/www/components/Front/index.js
```diff
@@ -46,13 +49,27 @@ const styles = {
...
 }
 
-export const RenderFront = ({ isEditor, front, nodes }) => {
+export const RenderFront = ({ front, nodes }) => {
   const { t } = useTranslation()
+  const { isEditor, hasAccess } = useMe()
+  const { addAudioQueueItem, isAudioQueueAvailable } = useAudioQueue()
+  const { toggleAudioPlayer } = useAudioContext()
+
+  const showPlayButton = hasAccess && isAudioQueueAvailable
```
ovbm, line 58: shouldn't this be `isReadaloud` instead of `isAudioQueueAvaiblabe` ?

apps/publikator/lib/translations.json
```diff
@@ -652,6 +652,10 @@
       "key": "metaData/audio/source/kind/syntheticReadAloud",
       "value": "automatisch vorgelesen"
     },
+    {
+      "key": "metaData/audio/isReadAloud",
```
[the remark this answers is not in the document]
annatraussnig: done + tested

apps/www/components/Front/index.js
annatraussnig: for the front we use a query to get audio data that's only available if the queue is available … we might refactor this to put play buttons on the marketing front, but after the release.

PR #235, opened 2022-10-10
Publikator: Files

patrickvenetz:
This Pull Request allows to upload files and publish them.

Bildschirmfoto 2022-10-10 um 19 56 49

Flow to upload a file:
- Report imminent upload to API (`mutation repoFileUploadBegin`). Returns "upload URL".
- PUT request file as body to "upload URL".
- Report upload to be completed (`mutation repoFileUploadCommit`).

"upload URL" is a signed S3 URL. A file is uploaded to S3 directly from browser (as to avoid unnessary ingress/egress traffic).

Uses `axios` as fetch polyfill in Publikator because it provides progress events while uploading.

Uploaded files are private and require a signed URL to open. Signed URL will point to S3 directly.

You can "publish" a file to make it public (`mutation updateRepoFile`). It returns a valid (CDN/Asset) URL.

You can not unpublish a file.

- [x] Notice about not to upload sensitive data
- [ ] Check if internal links are linked in document (republik-assets.s3.eu-central-1.amazonaws.com)
- [ ] Install CORS on S3 bucket to allow posting to
- [ ] Update bucket policy (add `s3:GetObject`)


apps/publikator/components/Files/Details.js
```diff
@@ -0,0 +1,30 @@
...
+      <Label>
+        {displayDateTime(file.createdAt)}
+        {' · '}
+        {file.author?.name}
+        {isReady && !isPublic && (
+          <>
+            {' · '}
+            nicht öffentlich
+          </>
+        )}
```
annatraussnig, line 26: I'd use a config for this (see below).

apps/publikator/components/Files/Upload.js
```diff
@@ -0,0 +1,137 @@
...
+
+        await uploadCommit({
+          variables: { id },
+          refetchQueries: ['getFiles'],
+        })
+      }
+
+      setUploading(false)
+    } catch (e) {
+      setUploading(false)
```
annatraussnig, line 103: maybe print the error / show it somwhere?

apps/publikator/components/Files/Upload.js
```diff
@@ -0,0 +1,137 @@
...
+          data: file,
+          onUploadProgress: (progressEvent) => {
+            const { loaded: fileLoaded, total: fileTotal } = progressEvent
+
+            if (fileTotal) {
+              setProgress(
+                Math.round(((dropLoaded + fileLoaded) / dropTotal) * 100),
+              )
+
+              if (fileLoaded >= fileTotal) {
```
annatraussnig, line 81: how can the loaded portion be bigger than the total?

apps/publikator/components/Files/Upload.js
```diff
@@ -0,0 +1,137 @@
...
+                dropLoaded += fileLoaded
+              }
+            }
+          },
+        })
+
+        if (!(res.status >= 200 && res.status < 300)) {
```
annatraussnig, line 88: so th api error comes up here and the console error below, right? why do we exclude other error codes (40x, 500)?

apps/publikator/components/Files/index.js
```diff
@@ -0,0 +1,71 @@
...
+              {data?.repo?.files?.map((file) => (
+                <Tr key={file.id}>
+                  <Td>
+                    <File file={file} />
+                    <Details file={file} />
+                  </Td>
+                  <Td style={{ textAlign: 'right' }}>
+                    <Publish file={file} />
+                  </Td>
+                </Tr>
```
annatraussnig, line 61: this I'd extract in a FileRow component

packages/backend-modules/publikator/lib/File/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+const { ASSETS_SERVER_BASE_URL, AWS_S3_BUCKET } = process.env
+
+export function getKey(file: RepoFileRow): string {
+  return ['repos', file.repoId, 'files', file.id, file.name].join('/')
+}
+
+export function getSafeFileName(name: string): string {
+  const safeName = slugify(name)
+
+  if (!safeName?.length) {
```
annatraussnig, line 17: shouldn't we default to something (e.g. 'file_' + timestamp) instead of throwing?

annatraussnig (approved):
Looks pretty good – a few things I'd refactor, but none is a must.

annatraussnig:
Empty state looks a bit sad – a message (e.g. "No files yet.") would brighten it up a bit.
Screen Shot 2022-10-11 at 09 56 47

annatraussnig:
Last but not least: show warning on publish when a link is private.

apps/publikator/components/Files/Upload.js
patrickvenetz: It throws an error whenever code is outside 200 to 299. 4xx and 5xx are included.

packages/backend-modules/publikator/lib/File/utils.ts
patrickvenetz: Gets hard to recognize which file was which when you upload multiple files. Thus failing is intentional.

PR #238, opened 2022-10-22
fix(assets): pass necessary params to PDF endpoint, rest to asset server

### LINKED ISSUE #237: Erstellung eines A5 PDFs: "Invalid width"
### Bug Description

Es erscheint die Error Meldung "Invalid width" im Browser wenn man auf den Button A5 PDF generieren klickt.
grafik

### Expected Behaviour

Erwartetes Verhalten wäre die Anzeige des Artikels als PDF.

### Steps to reproduce

1. https://www.republik.ch/2022/10/06/einer-von-uns
2. PDF Button
3. A5 auswählen (egal ob mit oder ohne bild)
4. Öffnen

### Environment

Firefox

### Version

105.0.3

patrickvenetz:
`returnImage` expects size option to contain dimensions, not a string like "A4"

Fixes #237

### DIFF
--- a/packages/backend-modules/assets/express/pdf.js
+++ b/packages/backend-modules/assets/express/pdf.js
@@ -13,15 +13,26 @@ module.exports = (server) => {
       return res.status(403).end()
     }
 
-    // path with query
-    const path = req.originalUrl.replace(/^\/pdf/, '')
+    // get express path
+    const { path } = req.params
 
-    const result = await fetch(`${PDF_BASE_URL}${path}`, {
-      method: 'GET',
-    }).catch((error) => {
-      console.error('pdf fetch failed', { error })
+    // pick query parameters intended for PDF endpoint
+    // (pass along rest as options, later)
+    const { images, download, size, ...options } = req.query
+
+    // build URL to fetch PDF from
+    const fetchUrl = new URL(`${PDF_BASE_URL}/${path}`)
+
+    // add params
+    images && fetchUrl.searchParams.set('images', images)
+    download && fetchUrl.searchParams.set('download', download)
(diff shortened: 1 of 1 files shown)


packages/backend-modules/assets/express/pdf.js
```diff
@@ -13,15 +13,26 @@ module.exports = (server) => {
...
-      console.error('pdf fetch failed', { error })
+    // pick query parameters intended for PDF endpoint
+    // (pass along rest as options, later)
+    const { images, download, size, ...options } = req.query
+
+    // build URL to fetch PDF from
+    const fetchUrl = new URL(`${PDF_BASE_URL}/${path}`)
+
+    // add params
+    images && fetchUrl.searchParams.set('images', images)
```
trm217, line 27: Readability says "Big oof"

trm217 (approved):
Quite hard for me to predict the behaviour of the PDF endpoint.
Only issue I could see is values from the url query not being passed anymore. But I suppose you only passed those three params for a reason 👍🏻

patrickvenetz:
Yes. pdf rendering "engine" accepts only [download](https://github.com/orbiting/pdf/blob/main/src/render.js#L15), [images](https://github.com/orbiting/pdf/blob/main/src/render.js#L32) and [size](https://github.com/orbiting/pdf/blob/main/src/render.js#L33).

packages/backend-modules/assets/express/pdf.js
patrickvenetz: What would Readability make more comfortable? If clauses? map?

(Thought this might be more readable to frontend versed peepz.)

PR #240, opened 2022-10-24
Refactor Audio download

trm217:
Downloading a blob seems to causes issues in mobile browsers.
We're therefor reverting to a normal download link with a download property.

Programmatically create a anchor-tag and set the file as it's href as well as the download property. In most browsers, this will open the file in the same tab. Since cross-origin downloads are not permitted.
Due to this, this might behave inconsistent across browsers.
An other option would be to add the target="_blank" property, which would consistently open the file in a new tab. However this would not start the download in directly in any browser.


patrickvenetz (approved):
LGTM

PR #246, opened 2022-10-31
Setup Fixes

tpreusse:
Various small improvements discovered while using the platform for lobbywatch.

- remove republik and project r company names from setup scripts and example envs
- rm old sign in email template
- move maillog migration sql files to a existing module; otherwise it gets pruned away on deploy and can't run on a remote
- modernize screenshoter to be deployable again (on vercel)
- expose listed credential on community page when not signed in
- a mini fix to prevent explosions when no crowdfunding is active


tpreusse:
Tests fail because `SSG_DOCUMENTS_API_KEY` is not exposed to fork PRs. We could disable pre-rendering during build if it's missing but would recommend to just ignore for now.

packages/backend-modules/republik/migrations/sqls/20170724202357-migrate-crowdfunding-up.sql
```diff
@@ -159,9 +159,6 @@ CREATE TABLE companies(
   UNIQUE("name")
 );
 
-INSERT INTO companies("name") VALUES ('PROJECT_R');
```
patrickvenetz, line 162: Why did this go?

PR #248, opened 2022-11-01
feat(www/journal): redirect /journal to latest entry

jerstucki:
I've used the existing [middleware](https://nextjs.org/docs/advanced-features/middleware) to add the redirect, in order to avoid adding another page file _just_ for the redirect functionality (I assume it also has better performance characteristics since it runs earlier?).

Since this only runs on the server, there's no need to distinguish between logged in/out state; we can always get the latest entry using the `DocumentsApiKey`.

The custom Journal nav link can be removed, since the redirect is now handled in a single place.


trm217:
Thanks for your help on this!

It seems we can't use apollo from within the middleware due to the limited runtime of the middleware. 
If I remember correctly, the middleware runs in the [edge-runtime](https://nextjs.org/docs/api-reference/edge-runtime) where only certain browser- APIs are available. (no Node.js APIs are available in that runtime)
Many APIs such [eval](https://nextjs.org/docs/api-reference/edge-runtime) are not available from within the middleware. (which is the problem in this case, see the screenshot below)

image

With our initial usage of the middleware we had the same issue when using the [`jsonwebtoken`](https://www.npmjs.com/package/jsonwebtoken) library to validate the jwt. (we solved this by using [`jose`](https://www.npmjs.com/package/jose) instead)

I think it'd be best to move the logic into a separate page after all.
We could use SSG if we're fine with a redirect else we would have to use getServerSideProps instead.

jerstucki:
I see … a couple of thoughts:

- Are you even running in the edge runtime (isn’t that Vercel specific?)
- Also possible to just post a gql query string with `fetch`. Instantiating Apollo client for one-off requests like this on the server may be a bit overhead anyway?

Anyway, thanks for reviewing, just throwing out some ideas here. Feel free to ignore 😉

trm217:
Valid points:

- regarding the runtime, if there was a way to disable the runtime, I wouldn't be aware of it. (After all the shade the runtime initially got due to the limited APIs, I would expect to have heard about this 🤔).
- we could absolutely do it using fetch (we already do that to retrieve the users roles if no token is attached to the request)

trm217:
For some reason I can't edit the PR, I'll move merge it into an other working-branch and commit to main from there 👍🏻

trm217:
For some reason I can't edit the PR, I'll move merge it into an other working-branch and commit to main from there 👍🏻

PR #252, opened 2022-11-02
Fix importPayments if CAMT.053 file contains no entries

patrickvenetz:
A CAMT.053 statement file usually contains transactions, one or moore `Ntry` records – and also lacking any `Ntry` records.

This Pull Request handles that.


patrickvenetz:
Before shipping, need confirmation from finance, that on one account were actually no transactions.

fuenkchen (approved):

PR #254, opened 2022-11-04
fix(Audioplayer): background fix

ovbm:
set background to 'default' instdead of 'overlay', making the separation of the UI elements more clear. The reason why it was set to 'default' is because in the native apps in expanded mode, the status bar is colored 'default' and thus it is necessary to have the same color.

This PR sets 'overlay' as the default and only switches the background to 'default' when displayed in the native app and in expanded mode.

before:
Screenshot 2022-11-04 at 14 56 02

after:
Screenshot 2022-11-04 at 14 55 41


ovbm:
fixes: https://www.notion.so/republikmagazin/13-Audio-player-container-barely-visible-in-dark-mode-9d2238edb79d4256af9893765012295d

PR #258, opened 2022-11-08
Journal memo

annatraussnig:
This pull request introduces the amazing InlineMemo™ to the JournalEditor™. Whoot.

The entire memo logic was added to the styleguide and adapted to work with both editors. The dialog modal required some `useMe`-type identification, which was added to the `styleguide` as well.

I also moved the base `structure` and starter `slate value` for the journal to the styleguide, alongside `schema`. It seemed more logical there.

**New!** memo ponöpel in the toolbar:

https://user-images.githubusercontent.com/3907984/200813353-59acdae3-f149-47b5-95c4-1cc07c7add49.mov


packages/styleguide/src/components/Marker/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  drop: {
+    style: (isSelected) => ({ opacity: isSelected ? 0.6 : 0.3 }),
+    pickerColor: 'rgba(0,0,0,0.1)',
+  },
+}
+
+export const markerKeys = Object.keys(MarkerConfig)
+
+export const Marker: React.FC<{
+  marker: string
```
trm217, line 42: Any specific reason why this isn't typed? from what I see there is a limited choice of colors.

packages/styleguide/src/components/Editor/Forms/Memo/Picker.js
```diff
@@ -0,0 +1,39 @@
...
+    borderRadius: '4px',
+    marginRight: 10,
+    width: 32,
+    height: 32,
+    verticalAlign: 'middle',
+    cursor: 'pointer',
+  }),
+}
+
+export const Picker = ({ marker = 'yellow', isSelected, onClick }) => {
```
trm217, line 17: Could easily be TS I think, only special type would be the marker-prop I suppose (already commented on that not being typed in another component)

packages/styleguide/src/components/Marker/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      height: 3,
+      boxShadow: isSelected && 'red 0px 0px 2px',
+    }),
+  },
+  drop: {
+    style: (isSelected) => ({ opacity: isSelected ? 0.6 : 0.3 }),
+    pickerColor: 'rgba(0,0,0,0.1)',
+  },
+}
+
```
trm217, line 38: Marker Type could go here? 👀

packages/styleguide/src/components/Editor/Forms/Memo/MemoTree.js
```diff
@@ -101,6 +96,7 @@ const MemoContainer = ({ memo }) => {
 
   return (
     <CommentNode
+      CommentLink={({ children }) => children}
```
trm217, line 99: Shouldn't a fragment (`<>{children}</>`) be used here?

packages/styleguide/src/components/Editor/Core/helpers/text.ts
```diff
@@ -34,14 +36,26 @@ export const getCharCount = (nodes: (Descendant | Node)[]): number =>
...
+  node: NodeEntry<CustomAncestor>,
+): boolean => {
+  if (!node || !SlateElement.isElement(node[0])) return false
+  const formatText = elConfig[node[0].type].attrs?.formatText
+  if (!isTextInline(node[0])) return formatText
+  if (typeof formatText === 'boolean') return formatText
+  const parent = Editor.parent(editor, node[1])
+  if (!SlateElement.isElement(parent[0])) return false
+  return elConfig[parent[0].type].attrs?.formatText
+}
```
trm217, line 50: 😵‍💫 This could use some whitespace, looks like an abstract piece of art 😹

packages/styleguide/src/components/Editor/Core/Forms.tsx
```diff
@@ -46,35 +53,89 @@ const getForm = (
...
+// To handle 1) we start by going as far as specified in the config (until we bump into
+// an element with stopIteration flag set to true). Then we go down the tree from the root
+// node recontruct our array of forms, always in the same order.
+//
+// To handle 2) we check whether the top level form element is:
+//  - inline
+//  - has inline ancestor
+//  - has inline ancestor which only contains last element
+export const getForms = (editor: CustomEditor, path: number[]): FormData[] => {
+  if (!path || path === []) return []
```
trm217, line 130: Huch, what's the purpose of the second condition?

trm217 (approved):
I didn't quite understand all chances made in the config/elements. It tested the memo feature and it works as I would expect it to.
Therefore I assume it's all fine 👍🏻

packages/styleguide/src/components/Editor/Forms/Memo/Picker.js
annatraussnig: you are, of course, right

packages/styleguide/src/components/Marker/index.tsx
annatraussnig: yes!

packages/styleguide/src/components/Editor/Core/Forms.tsx
annatraussnig: Not sure if you refer to the condition in the code or the comments 😄 

`- has inline ancestor which only contains last element`
we only want to show a potential parent form if e.g. the memo fully overlaps with the link (if the memo is bigger than the link the writer can trigger the memo by clicking on the memo outside of the link. I rephrased for a bit more clarity.

`path === []`
this means the whole editor text is selected. It doesn't make sense to show forms in this case

PR #261, opened 2022-11-09
www: redirect /journal to latest journal-entry

trm217:
Continuation of #248. Thanks @jerstucki for your work! 🙇🏻


trm217:
🚢 -ing it!

PR #267, opened 2022-11-11
feat(www/audio): allow toggling playback in the actionbar, on the front and the marketing page carousel

trm217:
- feat(www/audio): add toggleAudioPlayback callback to play/pause the current track
- feat(www/actionbar): allow toggling playback in the action-bar
- feat(www/marketing): allow toggling-playback in the Article Carousel

## Todo
- [x] add function to check if documented equals the current active **_player_** item (instead of just saying the active-item is the one at the head of the queue)
  - [x] Move activePlayerItem state from AudioPlayerController to AudioProvider
  - [x] Replace checkIfActiveItem with checkIfActivePlayerItem in ActionBar and Marketing Carousel
- [ ] Pass toggleAudioPlayback and checkIfActivePlayerItem to the front-tiles
- [ ] check on how to prevent 1000+ checks against all tiles (perhaps using  Intersection Observer API?)


trm217:
The front integration causes lots of issues. All other integrations will be merged with #276. The front will be merged separately at a later time.

PR #270, opened 2022-11-15
Journal Share

annatraussnig:
### Share individual Journal Blocks

2 options:

- **share link to block**, which triggers a modal like plain old article share. The generated twitter/facebook card uses the title/description from the metadata, as well as a cropped screenshot of this block.

Screen Shot 2022-11-21 at 16 35 27

- **download full screenshot**. The screenshot is always in light mode, and has some branding in the footer, in case it is shared without linking back to the journal.

Republik-Journal(1)

### Notifications

Readers can now subscribe to the journal. And receive regular push/web notifications when we publish the journal with the notification option enabled.

Screen Shot 2022-11-22 at 11 16 24

Screen Shot 2022-11-28 at 14 31 50

The title of the notification "Journal: Abendaugabe jetzt online" can be adjusted in the metadata of the format page for the Journal. (Other formats can also be given custom notifications using the same method.)

### Also in this PR:

- resize journal image to max. 700px (max journal width)
- resize both light and dark mode image in one go (this should have been adjusted when we introduced the concept of dark mode images and simplifies the code in a few places)
- option to disable lazy loading for journal images (needed for the screenshots)


patrickvenetz:
Any chance to use query params or some other solution?

`/2022/11/08/journal/234` is bad for SE indexing and tracking:
- SE will penalize URLs with same content
- Tracking and reporting expects 1 URL = 1 document (or redirects)

e.g. a page that is not SSR, delivers headers but will redirect user to actual journal page?

annatraussnig:
query params are not an option with ssg.
idea number 2 is more promising: we could create a new page e.g. `/share/2022/11/08/journal/234` which redirects client side to `/2022/11/08/journal#234`. does that help though?

annatraussnig:
another solution for SE is to set the canonical URL as `/share/2022/11/08/journal/`
atm this is not the case but we could do it easier than any more involved solution. What do you think?

annatraussnig:
i changed the canonical, which should take of SE. Could we also use canonical url for tracking?

patrickvenetz:
You propably ran a little to far, a little too dirty.

I build [a proof-of-concept](https://github.com/republik/plattform/compare/journal-share...journal-share-poc-ssr?expand=1) which accepts query param and prints appropriate share image into header (for Twitter and others) to use.

Example with query:

Bildschirmfoto 2022-11-27 um 11 15 26

Example without query:

Bildschirmfoto 2022-11-27 um 11 20 14

Boils down to delegate `<Meta …>` "downwards".

If this is a viable option, I'd prefer this to anything else.

annatraussnig:
I didn't know about delegate, thank you so much!

apps/www/components/ActionBar/FlyerTileActionBar.js
```diff
@@ -0,0 +1,117 @@
...
+    true,
+  )
+
+  const downloadScreenshot = async (e) => {
+    e.preventDefault()
+    setDownloadLabel(t('article/actionbar/download/processing'))
+    trackEvent(['ActionBar', 'downloadJournalBlock', screenshotUrl])
+    const image = await fetch(shareImageUrl)
+    const imageBlog = await image.blob()
+    const imageURL = URL.createObjectURL(imageBlog)
```
patrickvenetz, line 79: Blobbing for download not so good (App won't handle it well?) and whole Spiel w/ "wird heruntergeladen..." could be kicked out.

We found [a decent approach for audio files](https://github.com/republik/plattform/blob/8229dd284b4720e9ddfebc9f5b49ec4e8c203b52/apps/www/components/Audio/helpers/DownloadAudioSource.ts#L9-L18). It would need that code snippet and assets server to add `Content-Disposition` header if `&download=1` is appended.

apps/www/components/Article/Page.js
```diff
@@ -268,6 +270,16 @@ const ArticlePage = ({
...
   const routerQuery = router.query
 
+  useEffect(() => {
+    if (share) {
+      const node = document.getElementById(share)
+
+      if (node) {
+        node.scrollIntoView()
+      }
+    }
```
patrickvenetz, line 280: Would this work?

```suggestion
    if (share) {
      // scroll into view if element exists
      document.getElementById(share)?.scrollIntoView()
    }
```

apps/www/components/Article/ShareImageFlyer.js
```diff
@@ -0,0 +1,108 @@
...
+
+const Branding = () => {
+  const isDesktop = useMediaQuery(mediaQueries.mUp)
+  return (
+    <div {...styles.branding}>
+      <div {...styles.brandingInner}>
+        <p {...styles.brandingText}>
+          Mehr Kurzes:
+          <br />
+          www.republik.ch/journal
```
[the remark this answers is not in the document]
annatraussnig: I don't quite see the point of putting the url in the translation keys. For `Mehr Kurzes` it's another story. I changed.

Issue #250, opened 2022-11-01
Use equal spacing in navigation bar


MorgeMoensch:
The spacing is not equal, especially when you look at the right side.

![Screenshot_2022-11-01-09-11-56-61_d71cdd200a3182fddf8710b46c5dcebc.jpg](https://user-images.githubusercontent.com/62468040/199188858-4b00b5d4-e8af-4ac3-98f8-ac8046aaa369.jpg)

On Android 12, Republik 2.2.0.

MorgeMoensch:
After the Klimalabor was added, Spacing now looks fine.

PR #274, opened 2022-11-16
feat(SecondaryNav): adds secondary nav to newsletter, articles and di…

ovbm:
…alog pages

With the search relocated to the Secondary Navigation, it would make sense to show it on articles and newsletter pages. As it scrolls out of view, the impact on the reading experience is minimal.


trm217:
Should we fix #250 with this as well?

apps/www/pages/dialog.js
```diff
@@ -235,16 +235,16 @@ const DialogPage = ({ serverContext }) => {
...
 
   return (
     <Frame
-      hasOverviewNav={!tab}
+      hasOverviewNav
       raw
       formatColor='primary'
-      stickySecondaryNav={!tab}
+      // Only sticky if on /dialog without any query-params
+      stickySecondaryNav={!tab && !id}
```
trm217, line 247: Added this (see comment above this line)

trm217 (approved):
Looks good to me 👍🏻 Pushing it to staging

PR #279, opened 2022-11-17
refactor: update turbo-repo commands to use non legacy command flags

trm217:
## Todo

- [x] Test if deployments still work as expected.
  - [x] www-app on staging
  - [x] publiaktor-app on staging
  - [x] admin-app on staging
  - [x] api-app on staging
  - [x] automatic styleguide deploy to heroku (running build.sh on local didn't work. but works fine when manually deploying to heroku)

## Commits

- docs: update docs about turbo-repo commands to use 'filter' instead of 'scope'
- refactor(build-script): update turbo-repo command to no longer use legacy flags


trm217:
Did a test deploy for the following deployments in the staging-area and all were successful:
- www
- publikator
- admin
- api
- styleguide-docs

PR #280, opened 2022-11-17
fix(statistics): find previous paths via redirections

patrickvenetz:
matomo/collect did not check redirections table to find old paths

It now checks for previous paths and merges records


fuenkchen (approved):
lgtm

PR #281, opened 2022-11-18
feat(www): allow toggling playback from the front

trm217:
- feat(www/audio): add toggleAudioPlayback callback to play/pause the current track
- feat(www/actionbar): allow toggling playback in the action-bar
- feat(www/marketing): allow toggling-playback in the Article Carousel
- refactor(www/audio): move activePlayerItem state into provider and remove legacy state
- refactor(www/useAudioQueue): rename checkIfActiveItem -> checkIfHeadOfQueue
- refactor(www/audio): use of new checkIfActiveQueueItem callback on front and action-bar play-toggle
- refactor(www/audio): readd state to handle legacy-playeritem
- fix(www/audio): ensure ActionBar doesn't call togglePlayer fro active item after pressing play
- fix(www/marketing): only expand player when pressing play on the item that is not already playing
- refactor(www/audio): memoize checkIfActivePlayerItem
- feat(styleguide/templates): attempt passing logic to show pause-button to front-tiles
- refactor(front): pass Audio-Playbutton as a render function to the front-template
- refactor(front): pass Audio-Playbutton as a render function to the front-template


ovbm (approved):
LGTM

PR #283, opened 2022-11-18
feat(Track): track audio download

ovbm:


trm217 (approved):
There was a change that I couldn't really understand in the ExpandedAudioPlayer. The 'isLastItem'-prop wasn't necessary anymore, so I removed it.

PR #285, opened 2022-11-21
fix(www/nav): render the back-button in the iOS app on all pages but the magazin-front

trm217:
With #274 hasOverviewNav is not passed on the page component. In the Header however, the backbutton has previously only been rendered, if hasOvervierNav=false.
With this PR we attempt to regain the previous behaviour by generalizing the rule to render the back-button for logged-in users when ever a logged in user is in the iOSApp on a page other than the magazin-front


ovbm:
Extended the pages where backbutton doesn't need to be rendered to all the pages that are in the overviewnav, mirroring past behavior. Tested with userAgents in firefox. Can be deployed imo.

annatraussnig (approved):
let's get this merged!

PR #286, opened 2022-11-22
fix(Assets): fixes rounding to round down on calculating crop size

ovbm:
Math.ceil rounds up, which could lead to a situation where the crop height would exceed the file size.

Using Math.floor fixes the issue. 

Example: Using the crop 0x18y100w82h on an image with the dimensions 2126x2661

it would try to extract:
top: Math.ceil((18 / 100) * 2126) = 478.9 => 479
height: Math.ceil((82 / 100) * 2126) = 2182.02 => 2183
479+2183 = 2662, which is 1 pixel higher than the image - > throws error

With Math.floor ensures that the crop is always lower than the total size.

### DIFF
--- a/packages/backend-modules/assets/lib/returnImage.js
+++ b/packages/backend-modules/assets/lib/returnImage.js
@@ -187,10 +187,10 @@ module.exports = async ({
 
       if (crop && size) {
         pipeline.extract({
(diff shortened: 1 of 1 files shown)


patrickvenetz (approved):
LGTM

PR #293, opened 2022-11-29
www Journal Page refactoring

annatraussnig:
- extract code from article page into `flyer/` folder
- split different components (e.g. flyer nav, flyer footer…) into different files
- add typescript
- reorder imports hierarchically
- types for `Typography/Flyer`

**No kitten should be harmed and no functionality should be affected by this refactoring**

### DIFF
--- a/apps/www/components/Article/Page.js
+++ b/apps/www/components/Article/Page.js
@@ -66,7 +66,7 @@ import SectionFeed from '../Sections/SinglePageFeed'
 import HrefLink from '../Link/Href'
 import { withMarkAsReadMutation } from '../Notifications/enhancers'
 import ShareImageFlyer from '../Flyer/ShareImage'
-import FlyerPage from '../Flyer'
+import Flyer from '../Flyer'
 
 import { getMetaData, runMetaFromQuery } from './metadata'
 import ActionBarOverlay from './ActionBarOverlay'
@@ -656,7 +656,7 @@ const ArticlePage = ({
                 </div>
               )}
               {isFlyer ? (
-                <FlyerPage
+                <Flyer
                   meta={meta}
                   documentId={documentId}
                   inNativeApp={inNativeApp}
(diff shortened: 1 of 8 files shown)


apps/www/components/Article/Page.js
```diff
@@ -69,20 +65,21 @@ import SectionNav from '../Sections/SinglePageNav'
 import SectionFeed from '../Sections/SinglePageFeed'
 import HrefLink from '../Link/Href'
 import { withMarkAsReadMutation } from '../Notifications/enhancers'
-import { cleanAsPath } from '../../lib/utils/link'
-import { getMetaData, runMetaFromQuery } from './metadata'
-import FlyerFooter, { FlyerMeta, FlyerNav } from './Flyer'
-import ShareImageFlyer from './ShareImageFlyer'
-import { getFlyerTileActionBar } from '../ActionBar/FlyerTileActionBar'
+import ShareImageFlyer from '../Flyer/ShareImage'
+import FlyerPage from '../Flyer'
```
ovbm, line 69: I find it confusing that the import form index is called FlyerPage, I'd expect it to be called Flyer

packages/styleguide/src/components/Typography/Flyer.tsx
```diff
@@ -283,10 +282,10 @@ const listItemRule = css({
   },
 })
 export const ListItem: React.FC<{
-  attributes: any
+  attributes?: any
   children: React.ReactNode
-  style: any
-}> = ({ children, attributes = {}, style = {} }) => {
+  style?: any
```
ovbm, line 287: type css would make more sense here, no?

ovbm (approved):
LGTM

apps/www/components/Article/Page.js
annatraussnig: done

packages/styleguide/src/components/Typography/Flyer.tsx
annatraussnig: done

PR #295, opened 2022-11-29
Journal: block title as metadata title

annatraussnig:
Sharing of individual journal blocks: when the block has a title, this becomes the title on social media, and we use a default sentence as description.
If the block doesn't have a title, we keep using the title and description set for the entire journal.

Screen Shot 2022-11-29 at 17 08 40


patrickvenetz:
Maybe prepent title with meta.format.title or Journal: or Republik-Journal:

patrickvenetz (approved):
Other than prepending format title: LGTM

Gorgeous work as always.

PR #297, opened 2022-11-30
Stoerer for the journal

annatraussnig:
This PR integrates a paynote into the journal in 2(.5) cases:

### Case 1: no abo

Reader is not logged in -> we suggest a trial abo

Screen Shot 2022-12-12 at 17 21 49

### Case 2: during trial

Reader is logged in with a trial abo -> we suggest a real abo $$$

Screen Shot 2022-12-12 at 16 58 46

### Case 2.5: during trial in iOS native

We are not allowed to sell stuff in iOS -> we show a thank you note (consistent with article paynote)

Screen Shot 2022-12-12 at 16 59 10

The paynote either comes right after the first journal block, or (if a specific block ID is specified as `share` URL param) right after the shared block.

**Issue:** in the [...path].tsx page, a random `payNoteSeed` is generated, which we then use in the article page to pick a paynote variation and ensure that it doesn't hiccup to another variation when react rerenders. However, it occasionally does flicker from one variation to another. I would assume there is some misconfiguration of the `payNoteSeed` at play – but I am not entirely sure. An expert's eye would be much appreciated.
Also, paynoteSeed is not initialised in SSR mode. Is there a better place for it? **For now I put it in the Render component wrapped with useMemo.**

### DIFF
--- a/apps/www/components/Auth/CodeAuthorization.js
+++ b/apps/www/components/Auth/CodeAuthorization.js
@@ -52,6 +52,7 @@ const CodeAuthorization = ({
   authorizeSession,
   me,
   onSuccess,
+  black,
 }) => {
   const [code, setCode] = useState('')
   const [payload, setPayload] = useState('')
--- a/apps/www/components/Auth/CodeAuthorization.js
+++ b/apps/www/components/Auth/CodeAuthorization.js
@@ -6,7 +6,8 @@ import {
   Button,
   Interaction,
   Field,
-  A,
+  linkRule,
+  plainLinkRule,
   InlineSpinner,
   useColorContext,
 } from '@project-r/styleguide'
@@ -117,9 +118,13 @@ const CodeAuthorization = ({
   }, [payload])
 
   return (
-    <form onSubmit={onSubmit} ref={formRef}>
+    <form
+      onSubmit={onSubmit}
+      ref={formRef}
+      {...colorScheme.set('color', 'text')}
+    >
       {minimal ? (
-        <P>
+        <p>
           {t.elements('Auth/CodeAuthorization/description', {
             emphasis: (
               <Emphasis key='emphasis'>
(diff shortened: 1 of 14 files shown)


apps/www/components/Flyer/index.tsx
```diff
@@ -29,6 +31,37 @@ export type MetaProps = {
...
+)
+
+const RenderWithPaynote: React.FC<{
+  value: CustomDescendant[]
+  tileId?: string
+  repoId: string
+  documentId: string
+}> = ({ value, tileId, repoId, documentId }) => {
+  const seed = useMemo(() => Math.random(), [])
+  let idx = 2
```
trm217, line 52: Why is this default-value `2`? A comment could be helpful for the readability 👀

apps/www/components/Auth/CodeAuthorization.js
```diff
@@ -116,9 +116,13 @@ const CodeAuthorization = ({
...
   return (
-    <form onSubmit={onSubmit} ref={formRef}>
+    <form
+      onSubmit={onSubmit}
+      ref={formRef}
+      {...colorScheme.set('color', 'text')}
+    >
       {minimal ? (
-        <P>
+        <p>
```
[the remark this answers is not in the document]
annatraussnig: Yeah… this component renders a paragraph that explains what to do with the validation code. Originally it used `Interaction.P` for this, but this really doesn't work with `Flyer.P` we use elsewhere on this page. Since we use the default `p` directly below the form, I thought this might work to make the trial form usable both in article and flyer context.

Here is how it looks now:
Screen Shot 2022-12-13 at 09 24 26

Compare with using `Interaction.P`:
Screen Shot 2022-12-13 at 09 25 18

apps/www/components/Flyer/index.tsx
annatraussnig: bhahahaha vade retro magic number!!!

PR #300, opened 2022-11-30
Journal Paynote Link

annatraussnig:
_Raison d'être: restrict access to linked republik pages in the journal, so that only authorised readers can navigate to those pages._

**In Publikator:** We expose a new meta prop`isRestricted`

Screen Shot 2022-12-19 at 13 44 32

**In the Backend:** If a document is called and the following conditions are met:

- `isRestricted` switched on, either in the document itself or in the document format
- reader is not authorised

…the API return an empty string as `href` for all the internal links.

**In www:** When an unauthorised reader clicks on such an empty link in the journal, a modal comes up and kindly suggests a probelesen:

Screen Shot 2022-12-19 at 13 31 34

Alternately, the reader can choose to sign in through the modal:

Screen Shot 2022-12-19 at 13 31 39

**In Styleguide:** a new property `mini` was added to the `Overlay component`. When used, it prevents the overlay from rendering full height:

Screen Shot 2022-12-19 at 13 43 35


apps/www/components/Flyer/LinkBlocker.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <Overlay mini onClose={onClose}>
+      <OverlayToolbar onClose={onClose} title={tabs} />
+      <OverlayBody>
+        {tab === 'register' && (
+          <TrialForm
+            payload={trackingPayload}
+            onSuccess={() => false}
+            minimal
+            titleBlockKey='journal'
```
patrickvenetz, line 68:
```suggestion
            titleBlockKey='flyer'
```

(Pondering, if many other parts are pointing towards flyer, if `flyer` would be a better key?)

packages/styleguide/src/components/Typography/Flyer.tsx
```diff
@@ -328,12 +328,9 @@ const A = forwardRef<HTMLAnchorElement, BaseProps>(
 
 export const Link: React.FC<any> = ({ children, href, ...props }) => {
   const { Link } = useRenderContext()
-  if (!href) {
-    return React.Children.only(children)
-  }
   return (
     <Link href={href} passHref>
-      <A href={href} {...props}>
+      <A href={href || '#restricted'} {...props}>
```
patrickvenetz, line 333: Is `#restricted` still necessary?

```suggestion
      <A href={href} {...props}>
```

apps/www/components/Trial/Form.js
```diff
@@ -323,10 +325,9 @@ const Form = (props) => {
                       )}
                     >
                       <ArrowForwardIcon
-                        style={{ cursor: 'pointer' }}
+                        style={{ cursor: 'pointer', color: '#fff' }}
```
patrickvenetz, line 328: Change works for dark mode …

Bildschirmfoto 2022-12-19 um 19 16 15

… but not for bright mode anymore.

Bildschirmfoto 2022-12-19 um 19 16 19

Not sure why you nailed that one.

Form seems to be good enough without it? Here an example from Probelesen-Page on prod.

Bildschirmfoto 2022-12-19 um 19 18 26

Bildschirmfoto 2022-12-19 um 19 18 21

packages/styleguide/src/components/Overlay/docs.md
```diff
@@ -110,3 +110,21 @@ responsive: Desktop small
...
+## Overlay `mini`
+
+```react
+noSource: true
+plain: true
+responsive: Mobile
+span: 2
+---
+  <OverlayRenderer mini isVisible onClose={() => {}} mini>
+    <OverlayToolbar title='Mini Overlay' onClose={() => undefined} />
```
patrickvenetz, line 123: On mobile, it looks like a call out now.

Other call outs do not have a close button.

Do you think, we should hide it?

Bildschirmfoto 2022-12-19 um 19 21 33

patrickvenetz (commented):
Pushed some changes to backends. Exceptional job, really 👍

Tested locally; found some oddities.

apps/www/components/Trial/Form.js
annatraussnig: Ah it was for the case with `background: primary` (black on green doesn't look great)
You are right, in all other cases it doesn't make a lick of sense. Will make it more specific.

packages/styleguide/src/components/Typography/Flyer.tsx
annatraussnig: Yeah… you are right. We don't need this. I just thought `#restricted` might make our DOM more clear, semantically.

apps/www/components/Flyer/LinkBlocker.tsx
annatraussnig: Right you are.

PR #301, opened 2022-12-01
fix(Flyer): dimensions of shareimage tweak

ovbm:
before:
Screenshot 2022-12-01 at 09 11 32

after:
Screenshot 2022-12-01 at 09 13 26


ovbm:
- [] tested on love

PR #302, opened 2022-12-05
feat(OverlayForm): add overflow hidden to OverlayForm

felixmichel:


annatraussnig (approved):
scrolly preview still does funny things, but otherwise this really improves the editing experience.

PR #305, opened 2022-12-07
feat: extend ArticlePage to allow for special coloring to be used with Klimalabor

annatraussnig:


apps/publikator/components/editor/modules/meta/ui.js
```diff
@@ -214,10 +214,16 @@ const MetaData = ({
...
-                  label: t('metaData/darkMode'),
-                }
+              ? [
+                  {
+                    key: 'darkMode',
+                    label: t('metaData/darkMode'),
+                  },
+                  {
+                    key: 'climate',
+                    label: 'Klimalabor',
```
trm217, line 224: Perhaps we should also create a translation string for this?

apps/www/components/Article/Page.js
```diff
@@ -266,6 +266,7 @@ const ArticlePage = ({
   const articleContent = article?.content
   const articleUnreadNotifications = article?.unreadNotifications
   const routerQuery = router.query
```
trm217, line 268: use !! since the isClimate prop on the frame expects a boolean (in case one of the ? returns undefined this will convert undefined -> false)

apps/www/components/Frame/index.js
```diff
@@ -128,60 +131,67 @@ const Frame = ({
...
-            dangerouslySetInnerHTML={{
-              __html: t('noscript'),
-            }}
-          />
-        </Box>
-      </noscript>
-      <div
-        {...(footer || inNativeApp ? styles.bodyGrowerContainer : undefined)}
+      <ColorHtmlBodyColors
+        colorsObject={isClimate ? colors.climateColors : colors}
```
trm217, line 135: The ColorHtmlBodyColors has colors as a fallback if undefined is provided, so we don't have to provide it by ourself here again.

apps/www/pages/klimalabor.tsx
```diff
@@ -0,0 +1,20 @@
...
+      <p>
+        Und hier kommt <span>Und hier dann ein gelber Balken dahiner</span>
+        viel Inhalt
+      </p>
+      <Button primary>Das ist ein Button</Button>
+    </Frame>
+  )
+}
+
+export default Klimalabor
```
trm217, line 20: Get rid of this page, or else with the next FE-deployment, people will be able to see this under `www.republik.ch/klimalabor`.

packages/styleguide/src/components/Colors/ColorContext.tsx
```diff
@@ -175,20 +201,20 @@ export const ColorHtmlBodyColors = ({ colorSchemeKey = 'auto' }) => {
...
-                  `html[data-user-color-scheme="light"], html[data-user-color-scheme="light"] body { background-color: ${colors.light.default}; color: ${colors.light.text}; }`,
+                  `html[data-user-color-scheme="light"], html[data-user-color-scheme="light"] body { background-color: ${colorsObject.light.default}; color: ${colors.light.text}; }`,
                 ].join('\n'),
                 `}`,
               ].join('\n')
-            : `html, body { background-color: ${colors[colorSchemeKey].default}; color: ${colors[colorSchemeKey].text}; }`,
+            : `html, body { background-color: ${colorsObject[colorSchemeKey].default}; color: ${colorsObject[colorSchemeKey].text}; }`,
       }}
     />
   )
```
trm217, line 220: I suppose these changes will be fine, but be sure to deploy on staging and double-check if everything works as expected in normal articles (check light and dark-mode) and pages that force dark-mode (for-example the lesha articles).

packages/styleguide/src/theme/colors.js
```diff
@@ -203,6 +203,18 @@ const colors = {
...
+      default: 'pink',
+      primary: 'blue',
+      primaryHover: 'yellow',
+    },
+    dark: {
+      default: 'purple',
+      primary: 'yellow',
+      primaryHover: 'darkblue',
+    },
+  },
```
trm217, line 217: I am not really fond of having it in the default colors object. However I also don't see a better place, so I guess it's fine…
Perhaps use the identical color values as the default theme until the final KlimaLabor colors are defined?

apps/www/components/Frame/index.js
trm217, line 135:
```suggestion
        colorsObject={isClimate ? colors.climateColors : undefined}
```
The ColorHtmlBodyColors has colors as a fallback if undefined is provided, so we don't have to provide it by ourselves here again.

apps/www/components/Frame/OptionalLocalColorContext.js
```diff
@@ -0,0 +1,14 @@
+import { ColorContextLocalExtension } from '@project-r/styleguide'
+
+const OptionalLocalColorContext = ({ children, ...props }) => {
+  if (props.localColorVariables) {
+    return (
+      <ColorContextLocalExtension localColors={props.localColorVariables}>
+        {children}
+      </ColorContextLocalExtension>
+    )
+  }
```
trm217, line 10:
```suggestion

/**
 * OptionalLocalColorContext is a wrapper that provides a extends the local color context
 * with the provided localColorVariables-prop if provided.
 */
const OptionalLocalColorContext = ({ children, localColorVariables }) => {
  if (localColorVariables) {
    return (
      <ColorContextLocalExtension localColors={localColorVariables}>
        {children}
      </ColorContextLocalExtension>
    )
  }
```

apps/www/components/Article/Page.js
```diff
@@ -266,6 +266,7 @@ const ArticlePage = ({
...
+  const isClimate = article?.content?.meta?.climate
```
trm217, line 269:
```suggestion
  const isClimate = !!article?.content?.meta?.climate
```
use '!!' to convert possible undefined to false, since the isClimate prop on the frame expects a boolean.

apps/publikator/components/editor/modules/meta/ui.js
trm217, line 224: Perhaps we should also use a translation string here?

packages/styleguide/src/theme/colors.js
trm217: @ovbm What are your thoughts on this?

trm217 (changes requested):
See comments above ☝🏻

trm217 (commented):
Get rid of translation-string that aren't needed and we should be fine.

Tested it on staging as well and does not seem to break anything.
However, what seems odd to me, is that the header color is now also changed on clima-lab pages… I don't think that we really want that 🤔

trm217 (approved):

PR #306, opened 2022-12-07
feat: allow reporting comments for guests

trm217:
## Description

Adaption in the backend to allow for non logged-in users to report comments.
In the Slack notificaiton "Guest" is displayed instead of the users name.

In the frontend a new hook was added, which syncs the reports made by guests to the local-storage in form of a serialized object of the shape { [commentId]: [reportedAt] }.
If the stored value contains a specific commentId, the report action is disabled for that comment. Once a locally stored report has exceeded it's live time of 30 days, the entry in the reports object  is deleted.

## Tasks
- [x] bug: multiple anonymous reports are only counted as a single report


ovbm (approved):
LGTM

packages/backend-modules/discussions/graphql/resolvers/_mutations/reportComment.js
```diff
@@ -36,9 +34,12 @@ module.exports = async (_, args, context) => {
     }
 
     const reports = existingComment.reports || []
-    if (reports.findIndex(({ userId }) => userId === me.id) < 0) {
+    const userId = me?.id ?? `guest-${new Date().toISOString()}`
```
patrickvenetz, line 37: Change request to prevent tedious SQL queries.

Setting `reports[].userId` to a string which can't be parsed as a valid UUID(v4) makes querying JSON data somewhat hellish.

Using a dedicated prop comes in handy: Since we use prop name `pseudonym` for anonymous or anonymized questionnaire answers, I jumped on that one.

`pseudonym` in answers is a UUID. Instead of `guest-{ts}`, place a [newly generated UUID would work](https://github.com/republik/plattform/blob/49c8a262459454509dde9e26deed22feb5423f06/packages/backend-modules/utils/uuidForObject.js#L2) in `pseudonym`. (`reports[].reportedAt` holds timestamp already.)

Thus:

a) Set `reports[].userId` as either undefined, null or UUID (`me.id`)
b) Set `reports[].pseudonym` as either undefined, null or `uuid()` if me.id is missing.
c) Find reports by checking for either `reports[].userId` or `reports[].pseudonym`

patrickvenetz (changes requested):

patrickvenetz (approved):
LGTM

PR #310, opened 2022-12-14
fix(Merci): remove long lead text in prolong merci

ovbm:
Screenshot 2022-12-14 at 17 37 39


annatraussnig (approved):
lgtm – title is still too long imo, but that's out of scope for now.

PR #299, opened 2022-11-30
Add journal checkbox in subscription list

annatraussnig:
query subscription data for the journal and display a checkbox for this format at the top of the notification checkbox list.

Screen Shot 2022-11-30 at 15 05 00

PR #311, opened 2022-12-16
feat: introduce Meta.suggestSubscription

patrickvenetz:
This Pull Requests introduces `Meta.suggestSubscription`. It describes whether a section and its format is suggest by default to users.

It queries sections and formats as is, but now with `meta { suggestSubscription }` and filters for that flag.

In return, env variable `ONBOARDING_SECTIONS_REPO_IDS` is not longer needed.

It replaces https://github.com/republik/plattform/pull/299

Examples using "Journal":

/konto/benachrichtigungen

Bildschirmfoto 2022-12-16 um 08 18 59

/einrichten

Bildschirmfoto 2022-12-16 um 08 19 48

/rubriken

Bildschirmfoto 2022-12-16 um 08 20 23

/suche

(section "In alle Kürze" won't appear since `feed` flag is disabled)

Bildschirmfoto 2022-12-16 um 08 20 07

Pull Request includes [fix(search): drop match_all query in builders](https://github.com/republik/plattform/commit/d8436a5197a97b4e391a53e0d0ae93b40f258358): `match_all` as a fallback can cause unwanted side-effects if filter var is falsy

Release checklist:

- [ ] Create (or reuse) section "In aller Kürze", set meta `suggestSubscription: true`, `feed: false`. Publish.
- [ ] Link section to "Journal". Publish.
- [ ] Add meta `suggestSubscription: true` to repo IDs in `ONBOARDING_SECTIONS_REPO_IDS`
- [ ] Deploy code
- [ ] Remove `ONBOARDING_SECTIONS_REPO_IDS` env variable

### DIFF
--- a/apps/www/components/Notifications/SubscribedDocuments.js
+++ b/apps/www/components/Notifications/SubscribedDocuments.js
@@ -12,7 +12,6 @@ import {
 import { css } from 'glamor'
 import SubscribeCheckbox from './SubscribeCheckbox'
 import withT from '../../lib/withT'
-import { ONBOARDING_SECTIONS_REPO_IDS } from '../../lib/constants'
 import { withMembership } from '../Auth/checkRoles'
 
 const styles = {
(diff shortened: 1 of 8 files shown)


annatraussnig (approved):
LGTM – tested with ❤️

PR #313, opened 2022-12-20
fix(ActionBar): flex-warp for seriesEpisode

ovbm:
Quick fix until we have a true action bar revamp


annatraussnig (approved):
code looks good – haven't tested it yet

PR #315, opened 2022-12-21
feat(Carousel): add play button as in front in article

annatraussnig:
Now they're **everywhere** 😱

Screen Shot 2022-12-21 at 10 52 25

Well, almost everywhere… the current implementation of `handleQueueItem` (in `apps/www/components/Audio/hooks/useAudioQueue.tsx`) falls back to adding the `AudioPlayerItem` directly to the queue if the person is not logged in. 

This doesn't work well with our `TeaserAudioPlayButton`, which current only receives `documentId` as prop. Possible solutions include: 

- passing enough of the document's metadata to the button so that the player can work without an API call
- querying the API to get the audio data based on just the documentId (I imagine the `getDocument` query could do the job)

Anyway, this is a bit bigger than anticipated, so I'm leaving @patrickvenetz and @trm217 decide on how they wish to proceed. For now this new feature only works for our logged in users.


apps/www/components/Article/Page.js
```diff
@@ -387,6 +391,8 @@ const ArticlePage = ({
     }
   }, [trialSignup])
 
+  const showPlayButton = !extract && hasAccess && isAudioQueueAvailable
```
trm217, line 394: What's the reason for not showing the play-button to non-members?

trm217 (commented):
Looks fine and seems to work as expected.
Hand off to @patrickvenetz

apps/www/components/Article/Page.js
annatraussnig: Currently the button doesn't work for logged-out users. (I described why above in the PR)

PR #331, opened 2023-01-10
fix(www/postcard): add max-height to postcard text field

felixmichel:
Bildschirmfoto 2023-01-10 um 15 42 20


annatraussnig (approved):
marvellous 🤩

PR #332, opened 2023-01-11
chore(api): update jsonwebtoken to 9.0.0

trm217:
I saw [this report](https://github.com/advisories/GHSA-27h2-hvpr-p74q) and since we use once of the affected versions of this lib.
Even though we don't use the affected feature on the api, I decided to upgrade the version anyways since it requires no changes,
since we are already fulfilling all requirements to [migrate from v8 to v9](https://github.com/auth0/node-jsonwebtoken/wiki/Migration-Notes:-v8-to-v9).
(The only upgrade-requirement that has an affect on the features we use, is the requirement regarding the used algorithm for singing the key. We already use ES256 though, so we are fine.)

(As of now, we don't use jwt.verify in the backend at all, the token is only verified in the frontend (where we are using a different library anyways))


patrickvenetz (approved):
LGTM

PR #334, opened 2023-01-12
fix: added check before removing member role granted by perks

lucianak:
Added check if the user was already a member before the grant of the role by a perk, if so the member role is not removed with the end of the perk

### DIFF
--- a/packages/backend-modules/access/lib/perks/accessWithRole.js
+++ b/packages/backend-modules/access/lib/perks/accessWithRole.js
@@ -23,17 +23,10 @@ const give = async (campaign, grant, recipient, settings, t, pgdb) => {
 }
 
 const revoke = async (grant, recipient, settings, pgdb) => {
-  let isRoleRevoked = false
-  if (settings.role === MEMBER_ROLE) {
-    isRoleRevoked = await removeMemberRole(
-      grant,
-      recipient,
-      findByRecipient,
-      pgdb,
-    )
-  } else {
-    isRoleRevoked = await removeRole(grant, recipient, pgdb, settings.role)
-  }
+  const isRoleRevoked =
+    settings.role === MEMBER_ROLE
+      ? await removeMemberRole(grant, recipient, findByRecipient, pgdb)
+      : await removeRole(grant, recipient, pgdb, settings.role)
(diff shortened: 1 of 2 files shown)


packages/backend-modules/access/lib/perks/accessWithRole.js
```diff
@@ -21,7 +23,17 @@ const give = async (campaign, grant, recipient, settings, t, pgdb) => {
...
+  if (settings.role === memberRole) {
+    isRoleRevoked = await removeMemberRole(
+      grant,
+      recipient,
+      findByRecipient,
+      pgdb,
+    )
+  } else {
+    isRoleRevoked = await removeRole(grant, recipient, pgdb, settings.role)
+  }
```
patrickvenetz, line 36: Style:  `let` vars are harder to keep track on who is mutating it. Using `const` vars and/or a more verbose pattern is usually easier to read.

```suggestion
  if (settings.role === memberRole) {
    const isRoleRevoked = await removeMemberRole(
      grant,
      recipient,
      findByRecipient,
      pgdb,
    )

    if (isRoleRevoked) {
      debug('revoke', {
        recipient: recipient.id,
        revokedRole: settings.role,
      })
      return {
        recipient: recipient.id,
        revokedRole: settings.role,
        eventLogExtend: `.${settings.role}`,
      }
    }
  }

  const isRoleRevoked = await removeRole(grant, recipient, pgdb, settings.role)
```

patrickvenetz (changes requested):
Lovely job!

packages/backend-modules/access/lib/perks/accessWithRole.js
lucianak: I agree, but don't you think it leads to a lot of code duplication in this case?

fuenkchen: You could also do it like that, couldn't you?

```
const isRoleRevoked =
    settings.role === MEMBER_ROLE
      ? await removeMemberRole(grant, recipient, findByRecipient, pgdb)
      : await removeRole(grant, recipient, pgdb, settings.role)

fuenkchen (approved):
LGTM

PR #335, opened 2023-01-13
Harden qry stmts, validate email inputs

patrickvenetz:
Avoids unprepared SQL statements and validates email from user inputs.


fuenkchen (approved):
LGTM

PR #338, opened 2023-01-19
refactor(www): enhance AssetsImage to load static files directly if in dev

trm217:
A oversight on my part while creating the asset-image.
As of now, images that are added to the static assets folder, can't be loaded by the AssetImage. (Once deployed to staging or production it works finde however.

With these changes, the AssetImage attempts to load images from the static assets folder directly with the next.js image-loader while in dev-mode.

### DIFF
--- a/apps/www/lib/images/AssetImage.tsx
+++ b/apps/www/lib/images/AssetImage.tsx
@@ -1,12 +1,32 @@
+import { useMemo } from 'react'
 import Image, { ImageProps } from 'next/image'
 import assetServerImageLoader from './assetServerImageLoader'
+import { isDev } from '../constants'
 
 /**
  * AssetImage is a wrapper around the Next.js Image component
  * that uses the asssets-server image-loader.
  */
-const AssetImage = (props: Omit<ImageProps, 'loader'>) => (
-  <Image {...props} loader={assetServerImageLoader} />
-)
+const AssetImage = (props: Omit<ImageProps, 'loader'>) => {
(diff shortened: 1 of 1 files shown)


apps/www/lib/images/AssetImage.tsx
```diff
@@ -1,12 +1,32 @@
...
-  <Image {...props} loader={assetServerImageLoader} />
-)
+const AssetImage = (props: Omit<ImageProps, 'loader'>) => {
+  // Local static images that are not yet on the asset-server
+  // can be loaded directly from the local file system while in development.
+  const loadStaticFromLocal = useMemo(() => {
+    if (
+      isDev &&
+      typeof props?.src === 'string' &&
+      props.src.startsWith('/static')
```
jerstucki, line 17: Nit: `props` is not optional, so no need for `?`?

Not sure if we should check for the other possible types (`StaticRequire | StaticImageData`) here too. I.e. `src.default.src` or `src.src` (oh my).

jerstucki: Also, can we just check for `/` instead of `/static`? If the image's at `public/foo.jpg` that would be the case (`/static` is not necessary anymore, no?)

jerstucki (approved):
👍

apps/www/lib/images/AssetImage.tsx
trm217: Will support StaticImageData but not StaticImageRequire as it is not exported from next anyways

PR #340, opened 2023-01-23
fix: add ? to questionnaire trying to fix the bug

felixmichel:


annatraussnig (approved):
This feels quite random, but I don't see a big negative in merging this change all same.

PR #343, opened 2023-01-30
Feat climatelab counter

felixmichel:
Add small numbers to climate lab counter, the small numbers are added through the props. There is an bool extended which hides the small numbers on the climate lab landing page.

Bildschirmfoto 2023-01-30 um 09 50 57

Bildschirmfoto 2023-01-30 um 09 51 22

### DIFF
--- a/apps/www/components/Climatelab/Counter.js
+++ b/apps/www/components/Climatelab/Counter.js
@@ -5,26 +5,70 @@ import { css } from 'glamor'
 import {
   mediaQueries,
   fontStyles,
-  P,
   Loader,
   useColorContext,
+  convertStyleToRem,
+  Label,
+  Editorial,
+  createFormatter,
 } from '@project-r/styleguide'
 
 import { gql } from '@apollo/client'
-import { countFormat } from '../../lib/utils/format'
+import { countFormat, swissNumbers } from '../../lib/utils/format'
 
 const styles = {
-  container: css({}),
+  container: css({
+    marginTop: '20px',
+  }),
+  primaryNumberContainer: css({
+    marginBottom: -20,
+    [mediaQueries.mUp]: {
+      marginBottom: 0,
+    },
+  }),
   primaryNumber: css({
     display: 'block',
-    marginBottom: -6,
-    fontSize: 80,
+    marginBottom: -15,
+    fontSize: 90,
     ...fontStyles.serifTitle,
     lineHeight: 1,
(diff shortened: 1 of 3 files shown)


apps/www/components/Climatelab/LandingPage/Page.tsx
```diff
@@ -76,9 +76,9 @@ const LandingPage = () => {
             {...css({ marginTop: 111, [mediaQueries.mUp]: { marginTop: 162 } })}
           >
             <Counter />
-            <p {...styles.text}>
+            {/* <p {...styles.text}>
```
annatraussnig, line 79: This should go upon release, right?

apps/www/components/Climatelab/config.js
```diff
@@ -2,6 +2,7 @@ export const climateColors = {
   light: {
     default: '#5A47E1',
     text: '#FFFFFF',
+    textSoft: '#FFFFFF',
```
annatraussnig, line 5: are you sure it makes sense to set `textSoft` to be the same as `text`? Maybe we can just keep using the `text` color, then?

apps/www/components/Climatelab/Counter.js
```diff
@@ -36,8 +75,16 @@ const query = gql`
   }
 `
 
-const Counter = ({ data }) => {
+const Counter = ({
+  data,
+  smallNumbers,
```
annatraussnig, line 80: I am very confused `smallNumbers` to be a weird name. Small in relation to what? Is there a bigNumbers hiding somewhere else? 
(After looking at publikator...)
Suggestion: How about `statisticsBreakout`, or `demographics` or something like that? And `showStatistics` or `showDemographics` instead of `extended`?

annatraussnig (approved):
Some suggestions for clarity's sake

apps/www/components/Climatelab/config.js
felixmichel: outside of the climate lab landing page I use the real textSoft colour

apps/www/components/Climatelab/Counter.js
felixmichel: sounds good, will do the changes

apps/www/components/Climatelab/LandingPage/Page.tsx
felixmichel: yes, i can delete those lines

PR #347, opened 2023-02-07
Discussion respecing allowed roles

fuenkchen:
This PR:

- adds option in meta section of publikator to open up discussion (article or discussion) for climate users
- adds handling of allowedRoles when document is published
- adds script to add new access grant to all climate users which handles expiring of trial
- changes edit and voting mutations on discussions so that users with specified roles are allowed to do these actions
- changes special handling for box on account page

Can be tested on love:
- if you subscribe now to Klimalabor you should only get `climate` role
- e.g. in this discusssion you should be able to comment, vote, ... https://www.republik.love/2023/01/31/mit-voller-kraft-gegen-die-klimakrise-diskutieren-sie-mit/diskussion and and here you should not be able: https://www.republik.love/dialog?t=article&id=69ea384b-1cd9-4e83-876b-e721e0e82e09
- can be changed in publikator for each discussion

### DIFF
--- a/apps/publikator/components/editor/modules/meta/ui.js
+++ b/apps/publikator/components/editor/modules/meta/ui.js
@@ -272,6 +272,31 @@ const MetaData = ({
           black
           getWidth={() => '50%'}
         />
+        {['article', 'discussion'].includes(titleData.meta.template) && (
+          <>
+            <Checkbox
+              checked={node.data
+                .get('discussionAllowedRoles')
+                ?.includes('climate')}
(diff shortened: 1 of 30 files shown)


patrickvenetz:
121e3af22f012121fbe352d372b123bbc7d0b3a7 won't render checkbox if another discussion document is set

Bildschirmfoto 2023-02-09 um 17 38 11

Bildschirmfoto 2023-02-09 um 17 38 17

PR #349, opened 2023-02-08
fix(Translations): update stoerer

annatraussnig:


lucianak (approved):

trm217 (commented):
Why isn't that dynamic? After all, there is a query for it.

PR #359, opened 2023-02-15
feat: introduce 'packages/icons'

trm217:
## Description

This PR adds a new package to the mono-repo, that exports a set of SVG files as react.js components.
The reason for this new package is that we would like to control our icons ourselves instead of relying on a third-party package like `react-icons`.

### Changes to backend-module due to TS issues

`packages/backend-modules/auth/loaders/User.ts` was updated to filter out null values, since `transformUser` might return null, which in this case, isn't possible anyways.


trm217:
@jerstucki I think this is fine now as far as it goes for the setup.
The remaining task is replacing all the icons from react-icons, which will be a pain in the…
I'd however want to do that in a separate PR.

packages/icons/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "@vitejs/plugin-react": "^3.1.0",
+    "react": "^18.2.0",
+    "typescript": "^4.7.2",
+    "vite": "^4.1.0",
+    "vite-plugin-dts": "^1.7.3"
+  },
+  "peerDependencies": {
+    "react": ">=17.0.0"
+  }
+}
```
jerstucki, line 26: Not sure if this is still necessary in 2023 but would it be useful to add `"sideEffects": false` to indicate that the package is tree-shakeable?

packages/icons/vite.config.ts
```diff
@@ -0,0 +1,33 @@
...
+      // entry: entryObject,
+      entry: path.resolve(__dirname, 'lib/components/index.ts'),
+      name: 'RepublikIcons',
+      formats: ['es'],
+    },
+    rollupOptions: {
+      external: ['react', 'react-dom', 'react/jsx-runtime'],
+      output: {
+        globals: {
+          react: 'React',
```
jerstucki, line 26: I think we can remove globals if we only build ES?

packages/icons/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "name": "@republik/icons",
+  "version": "0.0.0",
+  "private": true,
+  "types": "dist/types/index.d.ts",
+  "main": "dist/icons.mjs",
+  "files": ["dist"],
+  "scripts": {
+    "generate:icons": "svgr --out-dir ./lib/components ./lib/svg",
+    "dev": "nodemon --watch lib/svg --exec 'yarn build'",
+    "build": "yarn generate:icons && tsc && vite build"
```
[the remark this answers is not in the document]
trm217: In regard to `tsc`, the default vite startet repo contains the the following as it's build command `"build": "tsc && vite build"`. I didn't second guessed it, and therefor left it as is. I will double check if it's needed.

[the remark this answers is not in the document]
trm217: In regards to `plugin-svgr`, from what I understood in the plugins docs, it is used to compile SVG files that are imported. As we don't import the SVG files themselves and instead just bundle a folder of svg icons, I don't the plugin matches our usecase.

packages/icons/vite.config.ts
trm217: I'd leave it in, as it doesn't hurt us if it's defined.
On the other way around, if we forgot about it, it wouldn't be great.
Therefor I'd keep it arround.

packages/icons/lib/icon-template.cjs
```diff
@@ -0,0 +1,13 @@
+// index-template.js
+let path = require('path')
+
+function defaultIndexTemplate(filePaths) {
+  const exportEntries = filePaths.map((filePath) => {
+    const basename = path.basename(filePath, path.extname(filePath))
+    const exportName = /^\d/.test(basename) ? `Svg${basename}` : basename
+    return `export { default as ${exportName}Icon } from './${basename}'`
```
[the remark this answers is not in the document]
trm217: Saw no option to enable this with the cli, there seems to be a option in the webpack plugin though.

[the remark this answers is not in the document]
trm217: Looked into it again, it's not possbile.

jerstucki, line 8: Mabe better to export icons as `Icon${basename}` rather than the other way around? Then the occasional `Svg` prefix 1 line above can be avoided.

packages/icons/.svgrrc.cjs
```diff
@@ -0,0 +1,36 @@
+// See https://react-svgr.com/docs/options/
+
+module.exports = {
+  icon: true,
+  typescript: true,
+  expandProps: true,
+  removeDimensions: false,
+  replaceAttrValues: {
+    "#000": "currentColor"
```
jerstucki, line 9: Would this also match `#000000`?

jerstucki (approved):
Just 1 small thing about the export names that might be easier to change now than later but otherwise LGTM 👍

packages/icons/.svgrrc.cjs
trm217: Good catch, I suppose I should add 'black' as well.

trm217: Nein! 🚢 😂

PR #362, opened 2023-02-15
chore: update turbo to v1.7.4

trm217:
- chore: update turbo to v1.7.4
- chore: make dev-task 'persistent'


jerstucki (approved):
Had to look up [what persistent means](https://turbo.build/repo/docs/reference/configuration#persistent) but it makes sense!

PR #363, opened 2023-02-16
fix(SenderPage): hide copy link icon in native app

jerstucki:
It doesn't seem work on Android, so we hide the copy to clipboard icon and just show the button to open the native share sheet.


trm217 (commented):
Fine to me, but this will hide it on iOS/iPadOS too.
If you use `inNativeApp` together with `const { isAndroid } = useUserAgent()`, we could just hide it for android.

jerstucki:
> Fine to me, but this will hide it on iOS/iPadOS too. If you use `inNativeApp` together with `const { isAndroid } = useUserAgent()`, we could just hide it for android.

Yes I know, but I think encouraging to use the native share sheet is better on both, so we don't really need it on iOS too.

PR #364, opened 2023-02-16
fix(Popover): limit height to 100dvh

jerstucki:
Ensure that even when we have lots of items in the popover the bottom part remains accessible and not hidden behind the browser toolbar.


jerstucki:
Doesn't work :(

PR #366, opened 2023-02-16
fix(Front): headline margins

jerstucki:
Add top margins on some Front headlines, so they don't appear glued to the top of the box if there's no text above them.


jerstucki:
Before/after:

one

two

three

With line above it's still the same:

four

patrickvenetz (approved):
LGTM

(Approved by AD, too)

patrickvenetz:
After deploy note: invalidate thumbails on pages like https://www.republik.ch/2023

PR #367, opened 2023-02-20
fix(discussions): check member role separatly

patrickvenetz:
This Pull Request revises `Discussion.userCanComment` prop.

It primarly checks if user has a role which is listed in `discussions.allowedRoles` (a). It wilfully ignores `member` role. `member` requires an additional check (b): user has to either have `debater` role or an active membership.

Revision fixes bug which allowed users with only `member` role to comment (without membership or `debater` role). Likely caused because array `DEFAULT_ROLES` was wrapped in array for filter (`![DEFAULT_ROLES].includes(role)`).

### DIFF
--- a/packages/backend-modules/discussions/graphql/resolvers/Discussion/userCanComment.js
+++ b/packages/backend-modules/discussions/graphql/resolvers/Discussion/userCanComment.js
@@ -3,7 +3,7 @@ const { hasUserActiveMembership } = require('@orbiting/backend-modules-utils')
 
 /**
  * User can comment when either:
- * - a) it has a role which is allowed to comment (discussion.allowedRoles)
+ * - a) they have a role which is allowed to comment (discussion.allowedRoles)
  * - b) it has member role and either debater role or an active membership
  */
 module.exports = async (discussion, _, context) => {
--- a/packages/backend-modules/discussions/graphql/resolvers/Discussion/userCanComment.js
+++ b/packages/backend-modules/discussions/graphql/resolvers/Discussion/userCanComment.js
@@ -4,7 +4,7 @@ const { hasUserActiveMembership } = require('@orbiting/backend-modules-utils')
 /**
  * User can comment when either:
  * - a) they have a role which is allowed to comment (discussion.allowedRoles)
- * - b) it has member role and either debater role or an active membership
+ * - b) they have the member role and either debater role or an active membership
  */
 module.exports = async (discussion, _, context) => {
   const { closed } = discussion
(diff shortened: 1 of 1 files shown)


packages/backend-modules/discussions/graphql/resolvers/Discussion/userCanComment.js
```diff
@@ -1,35 +1,33 @@
 const { Roles } = require('@orbiting/backend-modules-auth')
 const { hasUserActiveMembership } = require('@orbiting/backend-modules-utils')
 
-const DEFAULT_ROLES = ['member']
-
+/**
+ * User can comment when either:
+ * - a) it has a role which is allowed to comment (discussion.allowedRoles)
```
lucianak, line 6: being very pedantic here, but I think we shouldn't call user "it"
```suggestion
 * - a) they have a role which is allowed to comment (discussion.allowedRoles)
```

packages/backend-modules/discussions/graphql/resolvers/Discussion/userCanComment.js
```diff
@@ -1,35 +1,33 @@
...
+ * - b) it has member role and either debater role or an active membership
```
lucianak, line 7:
```suggestion
 * - b) they have the member role and either debater role or an active membership
```

packages/backend-modules/discussions/graphql/resolvers/Discussion/userCanComment.js
```diff
@@ -1,35 +1,33 @@
...
-  if (isInAllowedRoles) {
+  // a) User can comment, if it has some of the allowed roles but member
+  const allowedRoles = discussion.allowedRoles.filter((r) => r !== 'member')
+  if (Roles.userIsInRoles(user, allowedRoles)) {
     return true
   }
 
+  // b) User can comment, if discussion allows member role, user has member role
+  // and either debater role or an active membership.
+  const isMemberAllowed = !!discussion.allowedRoles.find((r) => r === 'member')
```
lucianak, line 26: Are there discussions where this isn't the case? Or do you mean that the discussion explicitly allows everyone with a member role (with or without active membership) to participate?

lucianak (approved):
Looks good, thanks, and the logic seems fine.

packages/backend-modules/discussions/graphql/resolvers/Discussion/userCanComment.js
patrickvenetz: Code allows for `discussions.allowedRoles` to be an empty array. If so, user should not be able to comment.

jerstucki: Nit: IMHO this could be written more clearly as `discussion.allowedRoles.some((r) => r === 'member'` instead of (mis-) using `find` and the double-falsification `!!`

patrickvenetz: True!!

PR #370, opened 2023-02-22
feat(future-campaign): update average and default value on PriceSlider component

trm217:
- refactor(future-campaign): style average price based on 'isAverage' value instead of 'isDefault'
- feat(future-campaign): set average-price to 105 CHF


jerstucki (approved):
LGTM!

trm217:
Now all we need is @ovbm to verify the average-price.

PR #371, opened 2023-02-24
feat(voting): add filter to query value char length for Text questions

patrickvenetz:
This Pull Request allows querying for length of Text answers.

To achieve this, we populate and index `resolved.answers.resolved.value.length` with char length if question type is `Text`.

We then convert array with objects `resolved.answers` to [nested documents](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html) which gives us the power to query multiple props at once: `questionId` _and_ `length`.

Now, we need to extend our GraphQL schema with a filter interface:

```gql
input SubmissionsFilterInput {
  ...
  answers: [
    { questionId: ID!, valueLength: { min: Int, max: Int } }
  ]
  ...
}
```

It will then query `questionId` and look for `valueLength.min` and  `valueLength.max`.

### DIFF
--- a/packages/backend-modules/search/lib/inserts/questionnairesubmission.js
+++ b/packages/backend-modules/search/lib/inserts/questionnairesubmission.js
@@ -32,11 +32,15 @@ async function transform(row) {
             )) ||
           answer.payload?.value
 
+        const length =
+          type === 'Text' && typeValue?.length ? typeValue.length : undefined
+
         return {
           ...answer,
           resolved: {
             value: {
               [type]: typeValue,
+              length,
             },
             question: {
               id,
--- a/packages/backend-modules/voting/graphql/schema-types.js
+++ b/packages/backend-modules/voting/graphql/schema-types.js
@@ -249,9 +249,12 @@ type Questionnaire {
   turnout: QuestionnaireTurnout
 
   submissions(
+    "Find search term in answer values"
     search: String
+    "Find specific value in answers"
     value: String
     first: Int
+    "Filter submissions"
     filters: SubmissionsFilterInput
     sort: SubmissionsSortInput
     before: String
@@ -260,22 +263,36 @@ type Questionnaire {
 }
 
 input SubmissionsFilterInput {
+  "Return only submission with this ID"
   id: ID
+
+  "Omit submission with this ID"
   not: ID
 
+  "Return only submissions with these IDs"
   submissionIds: [ID!]
+
+  "Omit submissions with these IDs"
   notSubmissionIds: [ID!]
 
+  "Return only submissions with these answered question IDs"
   answeredQuestionIds: [ID!] @deprecated(reason: "use \`answers\` instead")
+
+  "Return only submissions with one or more answers" 
   hasAnswers: Boolean @deprecated(reason: "use \`answers\` instead")
 
+  "Return only submissions with these answered questions"
   answers: [SubmissionFilterAnswer]
   
+  "Return submission by these user IDs"
   userIds: [ID!]
 }
 
 input SubmissionFilterAnswer {
+  "Question wich must be answered"
   questionId: ID!
+
+  "Expected amount of characters in answer given"
   valueLength: SubmissionFilterAnswerValueLength
 }
 
--- a/packages/backend-modules/voting/graphql/schema-types.js
+++ b/packages/backend-modules/voting/graphql/schema-types.js
@@ -266,15 +266,20 @@ input SubmissionsFilterInput {
   submissionIds: [ID!]
   notSubmissionIds: [ID!]
 
-  answeredQuestionIds: [ID!]
-  hasAnswers: Boolean @deprecated(reason: "use \`answeredQuestionIds\` instead")
+  answeredQuestionIds: [ID!] @deprecated(reason: "use \`answers\` instead")
+  hasAnswers: Boolean @deprecated(reason: "use \`answers\` instead")
 
-  valueLength: SubmissionsInputValueLength
+  answers: [SubmissionFilterAnswer]
   
   userIds: [ID!]
 }
 
-input SubmissionsInputValueLength {
+input SubmissionFilterAnswer {
+  questionId: ID!
+  valueLength: SubmissionFilterAnswerValueLength
+}
+
+input SubmissionFilterAnswerValueLength {
   gte: Int
   lte: Int
 }
--- a/packages/backend-modules/voting/graphql/schema-types.js
+++ b/packages/backend-modules/voting/graphql/schema-types.js
@@ -268,8 +268,15 @@ input SubmissionsFilterInput {
 
   answeredQuestionIds: [ID!]
   hasAnswers: Boolean @deprecated(reason: "use \`answeredQuestionIds\` instead")
+
+  valueLength: SubmissionsInputValueLength
   
   userIds: [ID!]
+}
+
+input SubmissionsInputValueLength {
+  gte: Int
+  lte: Int
 }
 
 input SubmissionsSortInput {
--- a/packages/backend-modules/voting/graphql/schema-types.js
+++ b/packages/backend-modules/voting/graphql/schema-types.js
@@ -280,8 +280,10 @@ input SubmissionFilterAnswer {
 }
 
 input SubmissionFilterAnswerValueLength {
-  gte: Int
-  lte: Int
+  "Expect a minimum amount of characters in answer given"
+  min: Int
+  "Expect a maximum amount of characters in answer given"
+  max: Int
 }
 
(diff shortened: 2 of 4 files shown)


packages/backend-modules/search/lib/inserts/questionnairesubmission.js
```diff
@@ -32,11 +32,15 @@ async function transform(row) {
             )) ||
           answer.payload?.value
 
+        const length =
+          type === 'Text' && typeValue?.length ? typeValue.length : undefined
```
jerstucki, line 36: Maybe too early to comment here but beware that length is not necessarily the correct character count. Not perfect but probably better to use `[...typeValue].length` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#description).

Also, if the length is `0` this will be undefined. Intentional?

jerstucki: Nevermind, this is not actually about string length 🤦

patrickvenetz: (There is no such thing as too early.)

`undefined` when length 0 is intentional.

packages/backend-modules/voting/lib/Submission.js
```diff
@@ -80,20 +78,41 @@ const createSubmissionsQuery = ({
     terms: { userId: filters.userIds },
   }
   const mustQuestionnaireId = questionnaireId && { term: { questionnaireId } }
+  const mustSubmissionId = filters?.id && { term: { id: filters.id } }
+  const mustNotSubmissionId = filters?.not && { term: { id: filters.not } }
```
lucianak, line 82: Both `mustNotSubmissionId` and `mustNotSubmiddionIds` are in `must_not` (line 194), but here the filter is already defined as `filters?.not`, wouldn't that be a double negation?

packages/backend-modules/voting/lib/Submission.js
```diff
@@ -117,38 +136,60 @@ const createSubmissionsQuery = ({
...
+      questionId,
+    }))
+    .concat(filters?.answers || [])
+
+  const mustHaveAnswers = expectedAnswers && {
     bool: {
-      must: filters?.answeredQuestionIds.map((id) => ({
-        term: { 'resolved.answers.questionId': id },
-      })),
+      must: expectedAnswers.map(({ questionId, valueLength }) => {
```
lucianak, line 155: Could this be written in a more readable way?

packages/backend-modules/voting/graphql/schema-types.js
```diff
@@ -260,18 +263,46 @@ type Questionnaire {
...
+  "Omit submissions with these IDs"
   notSubmissionIds: [ID!]
 
-  answeredQuestionIds: [ID!]
-  hasAnswers: Boolean @deprecated(reason: "use \`answeredQuestionIds\` instead")
+  "Return only submissions with these answered question IDs"
+  answeredQuestionIds: [ID!] @deprecated(reason: "use \`answers\` instead")
+
+  "Return only submissions with one or more answers" 
+  hasAnswers: Boolean @deprecated(reason: "use \`answers\` instead")
```
lucianak, line 282: Do the deprecated endpoints still work?

packages/backend-modules/voting/lib/Submission.js
```diff
@@ -117,38 +136,60 @@ const createSubmissionsQuery = ({
...
-  const mustNotSubmissionIds = filters?.notSubmissionIds?.length && {
-    terms: { id: filters.notSubmissionIds },
+    nested: {
+      path: 'resolved.answers',
+      query: { term: { 'resolved.answers.payload.value': value } },
+    },
   }
-  const mustAnsweredQuestionIds = filters?.answeredQuestionIds && {
+
+  const expectedAnswers = (filters?.answeredQuestionIds || [])
```
lucianak, line 147: Is this how you filter for only the answers to the questions you want, and not all answers in the answer sets which contain the question to be answered?

lucianak (commented):
Looks good to me :100:

packages/backend-modules/voting/lib/Submission.js
patrickvenetz: Oh, I see were you went there. `filters.not` is just a prop name, like `filters.foobar`. It has no impact on negation.

packages/backend-modules/voting/graphql/schema-types.js
patrickvenetz: Yes, deprecated end points still work.

packages/backend-modules/voting/lib/Submission.js
patrickvenetz: It uses array of strings – question Ids – and foreces them to an array of objects `{ questionId: '123123' }` – using shape of `SubmissionFilterAnswer`.

It then concatinates this array of objects with objects in `filters.answers`, again, in the shape of `SubmissionFilterAnswer`.

Example:

```graphql
filters: {
  answeredQuestionIds: ['123']
  answers: [
    { questionId: '456' },
    { questionId: '789', valueLength: { max: 100 } }
  ]
}
```

will become:

```graphql
[
    { questionId: '123' },
    { questionId: '456' },
    { questionId: '789', valueLength: { max: 100 } }
]
```

packages/backend-modules/voting/lib/Submission.js
patrickvenetz: Probably, yes. Really depends on what you'd consider more readable.

I find Query DSL of ElasticSearch is what makes it rather hard to read.

```
must: expectedAnswers.map(({ questionId, valueLength }) => { questionId, valueLength }
```

Example

```
must: [
  { questionId: '123', valueLength: { min: 10 } }, 
  { questionId: '456' }, 
  { questionId: '789', valueLength: { max: 100 } }
]
```

But it then shovels this into Query DSL pattern of booleans and must nodes.

Or is destructuring throwing you off? Like this you'd find more readable?

```
must: expectedAnswers.map((expectedAnswer) => {
  const { questionId, valueLength } = expectedAnswer
  ...
}
```

patrickvenetz: Would you find this more readable?

```js
  const mustHaveAnswers = expectedAnswers?.length && {
    bool: {
      must: expectedAnswers.map(createMustHaveNestedAnswer),
    },
  }
```

And then elswhere

```js
const createMustHaveNestedAnswer = ({ questionId, valueLength }) => ({
  nested: {
    path: 'resolved.answers',
    query: {
      bool: {
        must: [
          { term: { 'resolved.answers.questionId': questionId } },
          valueLength && {
            range: {
              'resolved.answers.resolved.value.length': {
                gte: Math.max(1, valueLength.min) || 1,
                lte: valueLength.max || undefined,
              },
            },
          },
        ].filter(Boolean),
      },
    },
  },
})
```

(Which is a bit like moving difficult part into another code section, making mapping part more readable, but not actual code.)

patrickvenetz:
@lucianak I'll merge an ship this on, but let this not hinder more conversations on readability.

packages/backend-modules/voting/lib/Submission.js
lucianak: Maybe I just need to understand Query DSL better... I think the destructured approach is good, would it be possible to separate where the `questionId` comes from, and where the `valueLength` comes from?

packages/backend-modules/voting/graphql/schema-types.js
lucianak: As `hasAnswers` is no longer in the must part of the query implementation, how does GraphQl translate this again from the schema to the code if I ask for the `hasAnswers` filter? (I think you explained this already but I can't find it for this example)

packages/backend-modules/voting/lib/Submission.js
lucianak: Maybe something like this: 

```
  const mustHaveAnswers = expectedAnswers?.length && {
    bool: {
      must: expectedAnswers.map(({ questionId, valueLength }) => {
        return {
          nested: {
            path: 'resolved.answers',
            query: {
              bool: {
                must: [
                  questionIdTerm,
                  valueLengtRange
                ].filter(Boolean),
              },
            },
          },
        }
      }),
    },
  }


```

packages/backend-modules/voting/graphql/schema-types.js
patrickvenetz: Deprecation notice is misleading. Since we switched to indexing anonymized submissions, `hasAnswers` has no effect anymore; it will always return submissions with answers – they either belong to users or a pseudonym.

PR #379, opened 2023-03-02
Klimafragebogen improvements

felixmichel:


jerstucki (approved):
LGTM!

The only thing I find quite annoying is the scroll animation when navigating between the overview and a single question (and especially back). Why – when I select "zurück zur Übersicht" can't I just be right there (ie scrollTo without animating)?

PR #380, opened 2023-03-02
fix(ClimateQuestionnaire): remove edit link from special color context

annatraussnig:
fix darkmode bug for the edit link in the overview page.
please test it because i didn't 🙈


felixmichel:
i already fixed that in another branch

PR #386, opened 2023-03-04
Docs Setup

jerstucki:
Adds a new top-level package based on [Nextra](https://nextra.site/) for the new docs.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -23,7 +23,6 @@
     "@republik/eslint-config-frontend": "*",
     "@types/node": "16",
     "@types/react": "^18.0.28",
-    "@types/react-dom": "^18.0.11",
-    "typescript": "^4.9.5"
+    "@types/react-dom": "^18.0.11"
   }
 }
--- a/docs/package.json
+++ b/docs/package.json
@@ -1,0 +1,29 @@
+{
+  "name": "@republik/docs",
+  "version": "0.0.1",
+  "private": true,
+  "workspaces": {
+    "nohoist": [
+      "next"
+    ]
+  },
+  "scripts": {
+    "dev": "next",
+    "build": "next build"
+  },
+  "dependencies": {
+    "next": "^13.2.3",
+    "nextra": "^2.2.17",
+    "nextra-theme-docs": "^2.2.17",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  },
+  "devDependencies": {
+    "@next/eslint-plugin-next": "^13.2.3",
+    "@republik/eslint-config-frontend": "*",
+    "@types/node": "16",
+    "@types/react": "^18.0.28",
+    "@types/react-dom": "^18.0.11",
+    "typescript": "^4.9.5"
+  }
+}
--- a/docs/package.json
+++ b/docs/package.json
@@ -4,7 +4,9 @@
   "private": true,
   "workspaces": {
     "nohoist": [
-      "next"
+      "next",
+      "react",
+      "react-dom"
     ]
   },
   "scripts": {
--- a/docs/.eslintrc.js
+++ b/docs/.eslintrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  extends: [
+    '@republik/eslint-config-frontend',
+    'plugin:@next/next/recommended',
+  ],
+}
--- a/docs/next-env.d.ts
+++ b/docs/next-env.d.ts
@@ -1,0 +1,5 @@
+/// <reference types="next" />
+/// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/basic-features/typescript for more information.
--- a/docs/next.config.mjs
+++ b/docs/next.config.mjs
@@ -1,0 +1,8 @@
+import nextra from 'nextra'
+
+const withNextra = nextra({
+  theme: 'nextra-theme-docs',
+  themeConfig: './theme.config.tsx',
+})
+
+export default withNextra({})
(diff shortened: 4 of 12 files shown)


jerstucki:
I'd like to merge this as-is to set up the deployment to Vercel as soon as possible and add more content and customization in following PRs. 

Took me a while to get the workspaces setup working, so I'd appreciate a second pair of 👀 on this. The main issue was to be that yarn couldn't properly deal with multiple React versions across different workspace packages (this is necessary because Nextra requires v18 but our other apps are still on v17). Things behaved inconsistently when doing `yarn add` vs. a subsequent `yarn`. I thought the solution would be to use the [nohoist](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/) config, so each package could have their own react version but this made the Styleguide unhappy. So, the current working solution is to only use `nohoist` in the docs package.

I also needed to add docs to the prune script, since it's on the top level.

docs/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "nextra-theme-docs": "^2.2.17",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  },
+  "devDependencies": {
+    "@next/eslint-plugin-next": "^13.2.3",
+    "@republik/eslint-config-frontend": "*",
+    "@types/node": "16",
+    "@types/react": "^18.0.28",
+    "@types/react-dom": "^18.0.11"
```
trm217, line 28: Should we perhaps defined the TS-version explicitly?

trm217 (approved):
Worked fine for me with the first install 👍🏻 
LGTM, about deploying to Vercel, do you see any real need for that just yet?
Unless we use it as a Storybook equivalent, I don't really see why previews etc. would be of interest.

docs/package.json
jerstucki: I tried but then I ran into the same Problem as you did in the icons package – stuff started breaking all over the place 😅 I guess I could have pinned 4.7.2, but that would be a bit pointless since your PR upgrades to 4.9 anyway?

patrickvenetz:
Looking forward to it!

Tried to run api, but failed miserably:

```
$ yarn dev --filter=@orbiting/api-app

xxx/republik/plattform/packages/backend-modules/search/lib/DocumentZones.js:2
const visit = require('unist-util-visit')
              ^

Error [ERR_REQUIRE_ESM]: require() of ES Module xxx/republik/plattform/node_modules/unist-util-visit/index.js from xxx/republik/plattform/packages/backend-modules/search/lib/DocumentZones.js not supported.
Instead change the require of index.js in xxx/republik/plattform/packages/backend-modules/search/lib/DocumentZones.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (xxx/republik/plattform/packages/backend-modules/search/lib/DocumentZones.js:2:15)
    at Object.<anonymous> (xxx/republik/plattform/packages/backend-modules/search/lib/Documents.js:33:5)
    at Object.<anonymous> (xxx/republik/plattform/packages/backend-modules/search/graphql/resolvers/_queries/search.js:12:5)
    at xxx/republik/plattform/packages/backend-modules/apollo-modules-node/index.js:36:42
    at Array.map (<anonymous>)
    at requireDirectory (xxx/republik/plattform/packages/backend-modules/apollo-modules-node/index.js:24:8)
    at loadModule (xxx/republik/plattform/packages/backend-modules/apollo-modules-node/index.js:67:18)
    at Object.<anonymous> (xxx/republik/plattform/packages/backend-modules/search/graphql/index.js:3:27)
    at Object.<anonymous> (xxx/republik/plattform/packages/backend-modules/search/index.js:2:12)
    at Object.<anonymous> (xxx/republik/plattform/apps/api/server.js:9:5)
    at Object.<anonymous> (xxx/republik/plattform/apps/api/dev.js:2:16) {
  code: 'ERR_REQUIRE_ESM'
}
```

Same for:

```
$ yarn dev --filter=@orbiting/api-app...
```

This is likely because it became an [ESM "only" module starting version 3](https://github.com/syntax-tree/unist-util-visit/releases/tag/3.0.0).

According to yarn lock, `unist-util-visit` gets two more versions installed, 3.1.0 and 4.1.2 (besides 1.4.1 and 2.0.3).

Cause is probably a lack of version declaration in packages/backend-modules/search/package.json. Oder backend modules use `"unist-util-visit": "^2.0.2"`.

Maybe add that version to packages/backend-modules/search/package.json and see if that will pin version to 2.x for API?

jerstucki:
> Unless we use it as a Storybook equivalent, I don't really see why previews etc. would be of interest.

I think we eventually might use it like that?

Anyway, there's no point creating this package without seeing it deployed somewhere. I prefer setting up deployment early, so we don't invest a lot into a particular setup before we can prove it works E2E (and without breaking anything else). Also, this way we can work on it in across multiple smaller PRs continuously over time rather than the one big PR to rule them all. Finally, I see it a as a proof-of-concept of Vercel (and Deploy Previews) for other parts too.

jerstucki:
> Tried to run api, but failed miserably:

Oh my … thanks for testing that! I figured if `yarn build` and CI runs, that means it worked 😕

patrickvenetz:
@jerstucki ESM error looks like a runtime error. CI tests for API is stuck with unit tests and sadly don't touch search package. No further comment 😏

jerstucki:
@patrickvenetz Yes, adding the missing `unist-util-visit` dep to the search package seems to have fixed it.

jerstucki:
@patrickvenetz Re-checked that it's possible to `yarn build` and `yarn dev` all the apps 👌

PR #388, opened 2023-03-06
Tweaking /teilen and /cockpit

fuenkchen:
Cockpit:
- Gets new 2023s ticks
- Looses annotation in subscription chart

Teilen:
- no ShareChart anymore on /teilen page
- component ShareChart is still there with an additional comment. (Maybe first we should have a strategy with removing content which could be potentially needed in the future, but open for suggestions. Just did not want to make bigger changes without consent of frontend heroines


lucianak (approved):
LGTM :)

annatraussnig:
lgtm :) (still need to test)

annatraussnig (approved):

PR #390, opened 2023-03-07
Index and search meta.contributors instead of meta.credits.children.url

patrickvenetz:
This Pull Request streamlines how we get document contributors.

Contributors is an array with contributors objects. A contributor objects contains a `name`, a reference to what someone contributed (`kind`) and then either a user ID or a resolved user.

Contributors are sourced from:
- Links in credits line: Link label might be a name and URL a user ID
- Anaylized credits line: Lexer tries to find names (it won't yield a user ID)
- Contributors defined in documents meta block: `meta.contributors[]` may already contain contributor objects

We  merge all these sourced contributor objects by name and contribution (`kind`).

Changes include:

- Replaces `Meta.contributors` with a compound of `doc.meta.contributors` and extracted contributors from credits line
- Refactors `Meta.contributors[].user` resolver
- Indexes contributor user Ids, drops `credits.children[].url`
- Queries `meta.contributors` instead of `meta.credits.children.url`
- Removes deprecated `Meta.authors` resolver

After deployment, reindexing document index is necessary.


packages/backend-modules/documents/graphql/schema-types.js
```diff
@@ -107,9 +107,8 @@ type Meta {
   shareBackgroundImageInverted: String
 
   credits: JSON
-
-  authors: [User!]! @deprecated(reason: "use \`Meta.contributors\` instead")
```
annatraussnig, line 111: is this something to worry about for the frontend? did you check whether `authors` is in use somewhere?

annatraussnig (approved):
Gorgeous work as usual. One question mark regarding required amends in the frontend/publikator (see my comment).

patrickvenetz:
@annatraussnig Your review is much appreciated.

PR #393, opened 2023-03-08
chore(config): move chromium lambda config

jerstucki:
… out of the way, so it doesn't conflict with other (future) Vercel deployments


jerstucki:
Aaand it works 😄 – the new docs can already be deployed 🎉 

image

PR #396, opened 2023-03-09
refactor: replace 'react-icons' with own icon package in internal apps

trm217:
## Description

Replaces the icons imported directly from 'react-icons' that were previously used in the admin & publikator app. Most items should look identical, but some might have changed a bit.

The SVG files are material design icons that were [downloaded](https://fonts.google.com/icons?icon.query=open) from the official page.

## Todo

- [ ] Double check that the icons work as they did before (visually)


trm217:
@annatraussnig Could you please review these changes for the publikator?
(I suppose just clicking through all the publikator sites/states would be sufficient)

trm217:
See: https://github.com/republik/plattform/pull/408

PR #407, opened 2023-03-16
feat(www/audio): integrate MediaSession API with web-audioplayer

trm217:
## Description

This PR adds a new hook 'useMediaSession' that connects a given AudioQueueItem with the MediaSession API.
This new hook is used in the AudioPlaybackElement component, that is used to render and control the audio-element
when using the web-audioplayer.

In addition to that the helper-methods used in the AudioCover component have been refactored in order to be usable
in the context of the MediaSession API when having to create the Artwork that is to be shown in the OS player UI. 

## Commits

- feat(www/audio): add preliminary integration for media session api in web player
- refactor(www): expose getImageCropURL from AudioCover
- feat(www/audio): harden media session api integration
- fix(www): update getResizeUrl to return size if no 'size' query-param is given
- feat(www/audio): add new cover fallback-image
- refactor(www/audio): update useMediaSession to pass seekOffset to forward & backward callback
- fix(www/audio): fix seekTo in useMediaSession
- feat(www/audio): get rid of seekTo integration for media-session-hook


apps/www/components/Audio/AudioPlayer/ui/AudioCover.tsx
```diff
@@ -27,6 +27,10 @@ const styles = {
 const getResizefromURL = (url, size) => {
   const imgURL = new URL(url)
   const sizeString = imgURL.searchParams.get('size')
+  if (!sizeString) {
```
trm217, line 30: Before this was added, this function caused an exception if the size query-param was not given. This seemed like the right way to handle that case.

apps/www/components/Audio/AudioPlayer/AudioPlaybackElement.tsx
```diff
@@ -222,6 +233,23 @@ const AudioPlaybackElement = ({
...
+      onPlay,
+      onPause,
+      onSeekForward: (seekOffset: number) => onForward(seekOffset),
+      onSeekBackward: (seekOffset: number) => onBackward(seekOffset),
+      onSkipToNext: () => onSkipToNext(),
+      onStop: () => onClose(),
+      onRetrievePlayerState: () => ({
+        currentTime: mediaRef.current?.currentTime || 0,
+        duration: mediaRef.current?.duration || 0,
+        playbackRate: mediaRef.current?.playbackRate || 1,
```
jerstucki, line 248:
```suggestion
        playbackRate: mediaRef.current?.playbackRate ?? 1,
```

apps/www/components/Audio/AudioPlayerController.tsx
```diff
@@ -23,6 +23,7 @@ import {
   AudioPlayerLocations,
   AudioPlayerActions,
 } from './types/AudioActionTracking'
+import { useMediaSession } from './hooks/useMediaSession'
```
jerstucki, line 26: Unused import.
```suggestion
```

apps/www/components/Audio/hooks/useMediaSession.tsx
```diff
@@ -0,0 +1,124 @@
...
+) {
+  const callbackRefs = useRef(callbacks)
+
+  useEffect(() => {
+    callbackRefs.current = callbacks
+  }, [callbacks])
+
+  const updatePlayerState = useCallback(() => {
+    if ('mediaSession' in navigator) {
+      const playerState = callbackRefs.current.onRetrievePlayerState()
```
jerstucki, line 39: Needs check if `current` is not undefined
```suggestion
      const playerState = callbackRefs.current?.onRetrievePlayerState()
```

apps/www/components/Audio/hooks/useMediaSession.tsx
```diff
@@ -0,0 +1,124 @@
...
+        getMediaImage(playerItem, 256),
+        getMediaImage(playerItem, 384),
+        getMediaImage(playerItem, 512),
+      ],
+    })
+
+    if (playerItem) {
+      mediaSession.playbackState = isPlaying ? 'playing' : 'paused'
+
+      if (callbackRefs?.current) {
```
jerstucki, line 71: `callbackRefs` can't be undefined (just `current`)

```suggestion
      if (callbackRefs.current) {
```

apps/www/components/Audio/hooks/useMediaSession.tsx
```diff
@@ -0,0 +1,124 @@
...
+          updatePlayerState()
+        })
+        mediaSession.setActionHandler('nexttrack', () => {
+          callbackRefs.current.onSkipToNext()
+          updatePlayerState()
+        })
+        mediaSession.setActionHandler('stop', () => {
+          callbackRefs.current.onStop()
+          updatePlayerState()
+        })
```
jerstucki, line 95: There are a bunch more actions like `previoustrack`. I assume we don't need them, but just double-checking.

apps/www/components/Audio/AudioPlayer/ui/AudioCover.tsx
```diff
@@ -69,3 +68,29 @@ const AudioCover = ({
...
 export default AudioCover
+
+/**
+ * Get a resized and cropped image url
+ * @param imageUrl image to crop
+ * @param size defines the width and height of the image
+ * @param audioCoverCrop crop parameters
+ * @returns image url with crop and resize parameters
+ */
+export function getImageCropURL(
```
jerstucki, line 79: Put this in a separate file? I don't think it's great to export random functions from component files (too easy to create circular dependencies).

apps/www/components/Audio/hooks/useMediaSession.tsx
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export function useMediaSession(
+  playerItem: AudioQueueItem | null,
+  { isPlaying, callbacks }: MediaSessionOptions,
+) {
+  const callbackRefs = useRef(callbacks)
+
+  useEffect(() => {
+    callbackRefs.current = callbacks
```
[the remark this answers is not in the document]
jerstucki: I'm just not sure about the pattern of mutating a ref inside a useEffect hook. Is there something I'm not aware? Are there some documented rules about when/where refs should be mutated?

apps/www/components/Audio/hooks/useMediaSession.tsx
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    return () => {
+      mediaSession.setActionHandler('play', null)
+      mediaSession.setActionHandler('pause', null)
+      mediaSession.setActionHandler('seekforward', null)
+      mediaSession.setActionHandler('seekbackward', null)
+      mediaSession.setActionHandler('nexttrack', null)
+    }
+  }, [JSON.stringify(playerItem), isPlaying, updatePlayerState])
```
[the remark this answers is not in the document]
trm217: If our code base makes it to a time, where JSON is no longer the standard without a rewrite, I'd be amazed 🤣

apps/www/components/Audio/hooks/useMediaSession.tsx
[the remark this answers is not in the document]
trm217: Yeah, you're right. Seems to play into the arguments provided here: https://react.dev/learn/you-might-not-need-an-effect

trm217:
@jerstucki Could you look over the last 5 commits? 👀

jerstucki (approved):
Nice! Good call adding a test for the image url 👍

PR #411, opened 2023-03-20
fix(www): add sharp as a dependency for image optimization in production

trm217:
Adds sharp as a production dependency to our main frontend. (See the explanation [here](https://nextjs.org/docs/messages/install-sharp))


trm217:
Not necessarily, using Next-Image (not AssetImage) will still use squoosh/sharp and not the assets-server.
(Can't change the default loader for all Next/Image usages in the Next.js version that we use.
But yeah we could just not add it and hope that no one uses next/image instead of AssetsImage. Should be fine too.

jerstucki (approved):

PR #424, opened 2023-03-30
chore: extract tsconfig into `packages/tsconfig`

trm217:


packages/tsconfig/nextjs.json
```diff
@@ -0,0 +1,30 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
jstcki, line 3: I don't think this is necessary or even used by next.js (the latest auto-generated tsconfig.json from docs doesn't have it at all)

jstcki (approved):
LGTM!

packages/tsconfig/nextjs.json
trm217: I believe that we had this for IE, unsure tbh.

jstcki: I'd just take it out, since the current default doesn't specify it anyway. I doubt it's used since transpilation happens with swc/babel

PR #426, opened 2023-03-30
feat(www): get rid of code related to future-campaign

trm217:


patrickvenetz:
Ping me before merging.

Since we spread /5-jahre-republik, /5-jahre-republik/[code] and /verstaerkung-holen, we ought to make sure no 404 pages are returned.

We have some Publikator Docs to catch user in the most friendly way and lead them towards offers or trial access.

Would need some redirects hard coded redirects.

patrickvenetz (approved):
LGTM

apps/www/components/Frame/Popover/UserNav.js
```diff
@@ -173,29 +173,7 @@ const UserNav = ({
...
-                          href='/teilen'
-                          currentPath={currentPath}
-                          closeHandler={closeHandler}
-                          large
-                        >
-                          {t('nav/share')}
-                        </NavLink>
-                      ) : null}
-                      {!me?.activeMembership && me.accessCampaigns.length > 0 && (
+                      {me.accessCampaigns.length > 0 && (
```
patrickvenetz, line 176:
```suggestion
                      {me?.accessCampaigns?.length > 0 && (
```

(me could be falsy, `accessCamapaings` could be null)

PR #429, opened 2023-03-31
fix(www): only upsert-media-progress to server if member

trm217:
## Description

I noticed that the API has lots of erros of invalid upsertMediaProgress mutations.
While we check if the user is a member before storing the progress on the server,
we don't in the client.


trm217:
@patrickvenetz What do you think about how this should be handled? Somehow it seems odd to me that we don't store the media-progress for non members, even though they have the option to opt-in in the settings 🤔

patrickvenetz:
Grand catch.

I see two options:

1. We could declare it a "paid" feature. (As it is already.) In turn, we'd have to add a notice about it next to the checkbox. It has the nasty implication if you lose access, you can't mutate your progress.
2. We could declare it a for-everyone feature. In turn, we'd have to losen role check in mutations `upsertDocumentProgress`, `upsertMediaProgress`. We'd gather more data points.

Any other viable alternatives that spring to mind?

Since it's hard for me to see the "unqiue selling point" qualities of option 1, I lean towards option 2.

trm217:
I'd vote for 2, but I think there might be other people who would have a opinion on this. (Com+ evt.?)

trm217:
@patrickvenetz I suppose it's ready now. Could you please look over the backend changes again?
I tested it today and it seems to be working as expected.
I will do a final test next Wednesday and release it afterwards if everything is looking allright.

patrickvenetz (changes requested):

trm217 (commented):
I suppose this would be fine now. Test it tomorrow? @patrickvenetz

patrickvenetz (approved):
LGTM

PR #431, opened 2023-03-31
Fix Publikator Crashing on Extra Title Content

jstcki:
When extra paragraphs are in the title section (e.g. when adding one through editing the source), normalization tries to [unwrap](https://github.com/ianstormtaylor/slate/blob/slate%400.31.8/docs/reference/slate/change.md#unwrapnodebykey) extra nodes which results in a invalid document that throws the editor into an infinite loop.

[Removing](https://github.com/ianstormtaylor/slate/blob/slate%400.31.8/docs/reference/slate/change.md#removenodebykey) extra nodes solves this. However, content (e.g. a valid credits paragraph) will be lost. Maybe there's a better solution like removing extra paragraphs _before_ the last one.

### DIFF
--- a/apps/publikator/components/editor/modules/title/index.js
+++ b/apps/publikator/components/editor/modules/title/index.js
@@ -147,7 +147,7 @@ export default ({ rule, subModules, TYPE }) => {
                 }
                 if (reason === 'child_unknown') {
                   if (index >= subModules.length) {
-                    change.unwrapNodeByKey(child.key)
+                    change.removeNodeByKey(child.key)
                   }
                 }
               },


jstcki:
For the record, I think this approach is a bit radical because it destroys content. Better would be to somehow merge paragraphs during normalization but I don't know if that's possible with Slate.

The normalize callback only kicks in on extra paragraphs, assuming that the 2nd is of type lead and the 3rd is of type credits. So if a paragraph is added _before_ the credits, it will be assumed that this is the credits paragraph and the actual credits are removed.

PR #435, opened 2023-04-03
feat(EdgeQuestion): bigger portrait pictures

felixmichel:


apps/www/components/Questionnaire/Submissions/Submission.js
```diff
@@ -115,6 +116,7 @@ export const SubmissionAuthor = ({
       {displayAuthor.profilePicture && (
         <img
           {...styles.headerPicture}
+          style={{ ...customStylePicture }}
```
annatraussnig, line 119:
```suggestion
          style={customStylePicture}
```

shouldn't this work too?

annatraussnig (approved):
looks good, one suggestion

PR #438, opened 2023-04-06
fix(): delete cached keys and properly merge userAttributes

lucianak:


lucianak:
Latest version works on love

PR #441, opened 2023-04-11
feat: allow reporting user profiles

trm217:
## Description

Adds a report-user functionality. On a user profile next to the share-icon a flag-icon was added that prompts to enter a reason for reporting the user. When a user is reported, it is sent to the internal slack-channel, where previously only reported comments where shown.

### Screenshots

image

### DIFF
--- a/packages/backend-modules/republik/graphql/resolvers/_mutations/reportUser.js
+++ b/packages/backend-modules/republik/graphql/resolvers/_mutations/reportUser.js
@@ -1,5 +1,7 @@
 const slack = require('../../../lib/slack')
 
+const MAX_REASON_LENGTH = 500
+
 module.exports = async (_, args, context) => {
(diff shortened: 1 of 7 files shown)


packages/backend-modules/republik/graphql/resolvers/_mutations/reportUser.js
```diff
@@ -0,0 +1,19 @@
+const slack = require('../../../lib/slack')
+
+module.exports = async (_, args, context) => {
+  const { user: me, t, loaders } = context
+  const { userId, reason } = args
```
patrickvenetz, line 5: Maybe limit `reason` in FE and backends

patrickvenetz (approved):
If anyone can report a user – signed in or not:

LGTM

PR #443, opened 2023-04-13
ci: improve pipelines by seperating jobs

trm217:
## Description

We chose to setup the turbo-pipeline, so that the test-pipeline relies on the build-pipeline. In hinsight this doesn't make much sense and needlessly slows down the ci as well as running the tests locally.

To improve on this I updated the turbo-repo pipelines and split the github-workflow into a build & test job.

### Blocker

In order for tests to work isolated from a build, we must ensure that the transpiled dependencies are transpiled when running jest-tests. (As of now, I see no way of achieving this, besides using native Next.js v13 module-transpile, therefor we're stuck with this until #397 is merged.


trm217:
Won't work until Next.js 13 is merged as we can't run transpile modules in the tests :(

PR #447, opened 2023-04-17
fix(Cockpit): remove growth-goal

ovbm:
Screenshot 2023-04-17 at 17 05 56


patrickvenetz (approved):
LGTM

And, as mentioned, bars will be missing on republik.ch/cockpit19 and republik.ch/crowdfunding which we put up with.

PR #448, opened 2023-04-18
feat: implement a opt-out consent for the auto-play feature of the audio-queue

trm217:
- feat(api): add 'AUDIO_AUTOPLAY_OPTOUT' to revokable consents
- feat(www): implement opting out from audioplayer autoplay


patrickvenetz:
Stalling temporarily. Autoplay setting shall be stored in user attributes. Must provide API therefor.

(Consent works technically, but is not intended for sät.)

PR #450, opened 2023-04-20
feat: react 18 upgrade

trm217:
- chore: bump React, ReactDom and @types libs to latest v18
- fix: remove deprecated React.ReactType
- fix(styleguide): get rid off typescript errors
- fix(www): fix typescript errors
- fix: get glamor working with react 18 in the frontends
- fix(www): type children correctly in FC usage


patrickvenetz (approved):
Did check-up on backends parts, should work nicely.

jstcki (approved):
LGTM.

But we've now got a couple versions of `@types/react(-dom)` v18 flying around. I don't think this is a problem if build/test works but I'd unify them into the top-level package.json and just remove from individual packages. Also there's a v17 version from testing-library (check with `yarn why @types/react`). But also fine later.

PR #454, opened 2023-04-24
Fix Server-Client Hydration Errors

jstcki:
Since the React 18 upgrade, certain hydration errors lead to an actual error in dev mode. Production is not directly affected, since the errors will just be logged to the console but hydration mismatch leads to unnecessary re-rendering of the whole page. Let's fix that.

### Notes
- I mainly checked all main pages of www and publikator, and had a quick look at admin (no issues there).
- Fixed all the critical hydration errors I could find, and some non-critical ones (e.g. style mismatches).
- I did fixed _some_ key warnings but not all of them. They are not critical at this point.
- There's a React key warning on front/TeaserMyMagazine that I can't seem to solve. All items do have unique keys but I still get the warning in the console.
- Probably best to review commit-by-commit

### DIFF
--- a/packages/styleguide/src/components/Figure/SwitchImage.js
+++ b/packages/styleguide/src/components/Figure/SwitchImage.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { useColorContext } from '../Colors/useColorContext'
 
-const SwitchImage = ({ src, srcSet, dark, alt, ...props }) => {
+const SwitchImage = ({ src, srcSet, dark, alt, maxWidth, ...props }) => {
   const [colorScheme] = useColorContext()
 
   return (
@@ -11,6 +11,7 @@ const SwitchImage = ({ src, srcSet, dark, alt, ...props }) => {
         src={src}
         srcSet={srcSet}
         alt={alt}
+        style={{ ...props.style, maxWidth }}
         {...colorScheme.set('display', dark ? 'displayLight' : 'block')}
       />
       {dark && (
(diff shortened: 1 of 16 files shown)


trm217:
I'll look at it tomorrow. We had many of these errors already before React18.

jstcki:
@trm217 Cool, thanks! Yes, the errors were all there before but React 18 surfaced them to an annoying level (which is good in a way, since we shouldn't really ignore them).

packages/styleguide/src/components/Figure/SwitchImage.js
```diff
@@ -11,6 +11,7 @@ const SwitchImage = ({ src, srcSet, dark, alt, ...props }) => {
         src={src}
         srcSet={srcSet}
         alt={alt}
+        style={{ ...props.style, maxWidth }}
```
trm217, line 14: If it wasn't applied previously (as maxWidth is no valid attribute), are we sure that everything is still behaving the same way as before?

trm217 (approved):
Looks good, just curious about the maxWidth prop.

packages/styleguide/src/components/Figure/SwitchImage.js
jstcki: If only we had something like tests that would catch this 😏 But yeah, I wondered too. I checked the current prod/love fronts by taking full-page screenshots.

I think it doesn't really matter because maxWidth is always absurdly high (> 1000). I'm not sure why it's used at all but we can look at that later.

trm217:
Looks fine 👍

PR #457, opened 2023-04-26
Search & replace

annatraussnig:
## Todos:

- [ ] Search and replace titles/descriptions in metadata
- [ ] Make metadata replace optional (e.g. in the case of adding a soft hyphen)
- [ ] Add soft-hyphens to UI (I imagine this will be one of the main use-cases) 
- [ ] Style sidebar button

### DIFF
--- a/apps/publikator/components/editor/Replace.tsx
+++ b/apps/publikator/components/editor/Replace.tsx
@@ -1,5 +1,5 @@
 import { css } from 'glamor'
-import { useState } from 'react'
+import React, { useState } from 'react'
 import {
   Overlay,
   OverlayToolbar,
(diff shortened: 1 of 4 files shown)


apps/publikator/components/editor/Replace.tsx
```diff
@@ -0,0 +1,167 @@
...
+      document.removeEventListener('keydown', keyDownHandler)
+    }
+  }, [])
+
+  const handleClick = (e) => {
+    console.log('handler called')
+    e.preventDefault()
+    const event = new KeyboardEvent('keydown', { key: 'a' })
+    document.dispatchEvent(event)
+  }
```
annatraussnig, line 106: @trm217 the function that just won't do anything right.

felixmichel (approved):

PR #458, opened 2023-04-28
chore: update jest & testing-library dependencies and use jest to run tests in styleguide

trm217:
## Description

This PR contains:

- Update jest & @testing-libarary/react across all repos to the latest version.
- Use `jest` to run tests in the styleguide.
- Refator styleguide/editor tests and make use of `act`


trm217 (commented):
I believe this is ready to be merged now, I don't see anything else that would be worth doing.

PR #466, opened 2023-05-05
feat(www): audioplayer autoplay refactoring

trm217:
- feat(www/audio): add auto-play toggle to player
- fix(www): don't throw error if custom cta-component was not found
- feat(www/audio): ensure secondary-controls don't change width on toggle & update translations
- refactor(www/audio): use setup function to prepare web audio-elem (analog to how it's implemented in the App player)
- fix(www/audio): fix react error due to missing key in EmptyQueue translation elem


patrickvenetz (approved):
LGTM

PR #468, opened 2023-05-09
🌆 Sunset Journal

annatraussnig:
übrigens:

- [x] calendar env variable
- [x] "in aller kürze" `suggestSubscription: false`
- [ ] move journal from "in aller kürze" to "archiv"


trm217:
`apps/www/lib/middleware/journalRedirect.ts` can be removed as well.

trm217 (approved):

PR #469, opened 2023-05-10
fix(www): wait for contactState change

trm217:
- fix(www): wait for contactState change
- fix(republik-crowdfundings): check email if pledge.user is provided


patrickvenetz (approved):
LGTM

PR #472, opened 2023-05-17
fix(admin): avoid error if no credential is listed

patrickvenetz:
If no credential was listed, filter failed [since `profileRole.id` was undefined](https://github.com/republik/plattform/compare/pae/fix-admin-credentials?expand=1#diff-7093a7ec393e1533dedcd0a574e05c4647bebbf3d8f415718febbaa8a2d23169L116).

Sort bubbles listed credentials to top if there are any.

Refactor to use guard clause.


jstcki (approved):
👍

PR #473, opened 2023-05-19
refactor(styleguide): goodbye catalog 🫡

trm217:
# Description

Get rid of catalog (previously used documentation library) in the styleguide and remove associated code. In addition, we also removed other unused code that was not removed while migrating to the monorepo

### DIFF
--- a/apps/publikator/package.json
+++ b/apps/publikator/package.json
@@ -113,6 +113,7 @@
     "topojson": "^3.0.2",
     "unified": "^7.0.0",
     "unist-util-visit": "^2.0.3",
+    "url": "^0.11.0",
     "validator": "^10.11.0"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -18043,6 +18043,14 @@ url@0.10.3:
     punycode "1.3.2"
     querystring "0.2.0"
 
+url@^0.11.0:
+  version "0.11.0"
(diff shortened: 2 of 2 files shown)


patrickvenetz:
@trm217 With [changes](https://github.com/republik/plattform/pull/473/commits/8afcf947ea7f9ac9687988520daba8fe45556f32) made by @jstcki, it seems to be building as expected.

@jstcki Vercel deployment is failing, should remove this?

Also packing and shipping styleguide to S3 is not working properly. First, anyone aware if this is still necessary?

jstcki:
> @jstcki Vercel deployment is failing, should remove this?

It's unrelated to this PR, I had changed the env variables. Should be back to working again.
 
> Also packing and shipping styleguide to S3 is not working properly. First, anyone aware if this is still necessary?

From what I can tell [from the workflow](https://github.com/republik/plattform/blob/63a41faa898abcbeae49219d7ecd00a97fd5e3ee/.github/workflows/styleguide.yml#L28-L36), it's used to upload a npm package to S3 without having to publish it to npm. I assume this was used pre-monorepo – or with the individual dynamic component repos – to test some pre-release styleguide things. As I understand, we haven't been using the styleguide as npm package for a while now, so I don't think we still need this going forward.

jstcki (approved):
@trm217 I added the missing `url` dependency to Publikator, now it works 🎉  Good to go from my side!

(Yes, I know, we should use the standard `URL` but that's a story for another day because `url` is also used in the `mdast-react-renderer` package which we'd have to update as well, test if it still works etc. etc.)

PR #474, opened 2023-05-22
fix(republik-crowdfundings): allow deleting users with claimed membership

lucianak:
…ships


packages/backend-modules/republik-crowdfundings/graphql/resolvers/_mutations/deleteUser.js
```diff
@@ -214,7 +216,12 @@ module.exports = async (_, args, context) => {
 
     // if the user doesn't have pledges, nor grants, nor candidacies we can delete everything,
     // otherwise we need to keep (firstName, lastName, address) for bookkeeping
-    if (!hasPledges && !hasGrants && !hasCandidacies) {
+    if (
+      !hasPledges &&
+      !hasGrants &&
+      !hasCandidacies &&
+      !hasClaimedMemberships
```
patrickvenetz, line 223: I think this could work.

Mutation returns a user entity if it has not been deleted. This requires `hasClaimedMemberships` check in [return function](https://github.com/republik/plattform/pull/474/files#diff-291b776f183fdfeb31183333852d279ef5135904be95929bf40f31b66cb7a0a1R259), too.

patrickvenetz (changes requested):

PR #475, opened 2023-05-22
Measure mp3 duration on commit

patrickvenetz:
This Pull Request measures mp3 duration on commit and adds it to meta data.

It will measure mp3 each time URL in `meta.audioSourceMp3` changes when committing and adds duration in miliseconds to `meta.audioSourceDurationMs`, or, if mp3 URL is not longer available, removes `meta.audioSourceDurationMs`.

Before this, mp3 duration was measured each time on publish, which was unncessary (audio source is deterministic unless URL pointing to audio file changes) and slowing publishing process down (re-indexing documents in case of an emergency has to download and measure each audio source).

It includes a script to migrate existing repos: Find latest commit, measure audio source duration, add it to meta, commit and, if applicable, publish.

**Pre deployment**

Remove `SUPPRESS_AUDIO_DURATION_MEASURE` env variable

**Post deployment**

```sh
heroku config:set SEARCH_PG_LISTENER=false -a api-republik-ch
```

```sh
$ heroku run bash -a api-republik-ch
# Wait until instance is spun up

$ DEBUG=publikator:lib:audioSource,publikator:script:migrateAudioSources node packages/backend-modules/publikator/script/migrateAudioSources.js
# This takes a while

$ DEBUG=search:lib:indexPgTable yarn pull:elasticsearch --index document repo --flush
# This takes a while, too
```

Remove repo, commit table records from `notifyTableChangeQueue`

```sh
heroku config:set SEARCH_PG_LISTENER=true -a api-republik-ch
```


packages/backend-modules/publikator/script/migrateAudioSources.ts
```diff
@@ -0,0 +1,200 @@
...
+          name: versionName,
+          meta: { script: 'migrationAudioSource' },
+          author,
+          scope: 'publication',
+          scheduledAt: now,
+          publishedAt: now,
+        })
+        debug('inserted milestone: %s (%s)', milestone.id, versionName)
+
+        await maybeDelcareMilestonePublished(milestone, tx)
```
lucianak, line 154: typo ;)

packages/backend-modules/publikator/script/migrateAudioSources.ts
```diff
@@ -0,0 +1,200 @@
...
+        createdAt: moment(workingCommit.createdAt).add(1, 'second'),
+      })
+      debug('inserted commit: %s', commit.id)
+      console.log(
+        'inserted commit',
+        repoId,
+        `https://publikator.republik.ch/repo/${repoId}/tree`,
+      )
+
+      if (isPublished) {
```
lucianak, line 117: Why do you need to replace the derivative, and can't just update the commitId?

lucianak (commented):
Looks good to me, only one thing: I think I would first update all meta.audioSourceDurationMs, and then use the standard publish mutation/code to update the already published ones with the new commit, to use the standard process and have less duplication.

patrickvenetz:
@lucianak Didn't understand your comment yet on what you would do? About duplication?

packages/backend-modules/publikator/script/migrateAudioSources.ts
patrickvenetz: Technically, we could just update the commit ID. But that would lead to some unwanted side-effects.

A derivative is "derived" from a commit. Could be a PDF version of a commit, or a synthetic read aloud audio file. An information or assets which belongs to a commit, but is appended asynchronously.

Once you publish a commit with derivates, they become available, too. (In fact, synthetic read aloud audio files are provided that way.)

If you were to add a commit, update its dervatives to that new commit and publish it it, it be working fine, too.

Now, if you or someone else decides to publish the previous commit again (aka. reverting to a previous version) its derivates would have been gone.

To prevent this, derivatives are copied if a commit was previously published.

PR #477, opened 2023-05-29
Replace geoip-database package w/ file from MaxMind

patrickvenetz:
This Pull Request replaces outdated "geoip-database" package with a GeoIP database file from [MaxMind](https://www.maxmind.com/). packages/backend-modules/auth/lib/geoForIP.js checks for GeoIP database file in `./local/geo/` and, if present, will look up an IP address and return city and country.

Change allows downloading a current version for MaxMind, using a MaxMind licensce key (`MAXMIND_LICENSE_KEY`) to (not tracked) folder `./local/geo/`.

File is downloaded when Heroku calls `heroku-postbuild` script after building "api" app.

A developer can download current GeoIP database file running `yarn geoip:download`.

"geoip-database" stopped updating when MaxMind introduced licensed access to GeoIP database file (= requires a payed account).

Outdated GeoIP database file resolved IP addresses to utterly faulty countries and cities, which in turn let to confusion on user side: City and country is shown when authenticating a session.


lucianak (commented):
lgtm :)

PR #480, opened 2023-05-31
Fix delete user with claimed membership

lucianak:


lucianak:
Fixed problem that was still there for canceled Memberships

PR #485, opened 2023-06-02
feat(www): add support for denying IPs

jstcki:


lucianak (changes requested):
env.example should be modified, otherwise LGTM :+1:

apps/www/server/index.js
```diff
@@ -47,6 +48,34 @@ app.prepare().then(() => {
     }),
   )
 
+  if (process.env.DENY_IPS) {
+    try {
+      // We use JSON.parse because to filter a IP range it needs to be specified as an array, e.g. [['127.0.0.1', '127.0.0.10']]
+      const denyIPs = process.env.DENY_IPS
+        ? JSON.parse(process.env.DENY_IPS)
+        : []
```
patrickvenetz, line 56: If there is no IP to deny requests for, setting up `server.use(ipfilter(...))` is probably not necessary?

apps/www/server/index.js
```diff
@@ -47,6 +48,34 @@ app.prepare().then(() => {
...
+        ipfilter(denyIPs, {
+          mode: 'deny',
+          logLevel: 'deny',
+          detectIp: (req, res) => {
+            const ip = req.headers['x-forwarded-for']
+              ? req.headers['x-forwarded-for'].split(',')[0]
+              : req.connection.remoteAddress
+
+            return ip
+          },
```
patrickvenetz, line 70: Built-in IP detected is likely more sane (uses [proxy-addr](https://www.npmjs.com/package/proxy-addr)). Would drop this.

In turn, if `!DEV`, set `trustProxy` to `true` since [it will run behind a proxy aka. Heroku Router](https://github.com/jetersen/express-ipfilter/blob/master/lib/ipfilter.js#LL38C7-L39C3).


```suggestion
          trustProxy: !DEV
```

jstcki: Ah yes, that should work. I just copied the example from the express-ipfilter README 🙄

apps/www/server/index.js
jstcki: Ah yes, actually the whole check was unnecessary because there's an `if` clause before.

lucianak (commented):
thank you for fixing it 🙌 looks good to me and I'd say let's leave it until we have a better solution

PR #496, opened 2023-06-15
fix(www): prevent audioplayer crashing the site if an invalid audio item is given

trm217:
- fix(www): add missing nullish-coal to AudioListItem
- fix(www): remove faulty audio-queue items if received by the API

### DIFF
--- a/apps/www/components/Audio/hooks/useAudioQueue.tsx
+++ b/apps/www/components/Audio/hooks/useAudioQueue.tsx
@@ -274,8 +274,15 @@ const useAudioQueue = (): {
     ? meWithAudioQueue?.me?.audioQueue ?? []
     : null
 
+  // In case faulty audio queue items are in the queue, remove them
+  resolvedQueue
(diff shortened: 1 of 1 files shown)


apps/www/components/Audio/hooks/useAudioQueue.tsx
```diff
@@ -274,8 +274,15 @@ const useAudioQueue = (): {
...
+  // In case faulty audio queue items are in the queue, remove them
+  resolvedQueue
+    ?.filter((item) => !item.document?.meta?.audioSource)
+    .forEach((item) => handleRemoveQueueItem(item.id))
+
   return {
-    audioQueue: resolvedQueue,
+    audioQueue: resolvedQueue?.filter(
+      (item) => item.document?.meta?.audioSource,
+    ),
```
jstcki, line 285: Just checking: audioQueue can be undefined here, that's intentional? (As opposed to an empty array …)

jstcki (approved):
1 question, but LGTM!

PR #497, opened 2023-06-15
feat(www): improve error-boundary

trm217:
# Fallback page
image


jstcki (changes requested):
Error Boundary Code and email content LGTM. 

But you probably should rebase, there is some unrelated (Turbo & AudioPlayer) code in this PR.

Suggestions for the page design:
- Remove first sentence, it feels a bit patronizing
- Move the "Seite neu laden" button up
- Vertically center the message

image

PR #500, opened 2023-06-20
Script Emails Politiker*innen-Fragebogen

lucianak:


packages/backend-modules/mail/script/sendMailsToSegment/politicianQuestionnaireMailing_06-2023.js
```diff
@@ -0,0 +1,193 @@
...
+  message: {
+    subject: 'Berichterstattung Nationalratswahlen - Unsere Fragen an Sie',
+    templateName: 'access_recipient_questionnaire_politician',
+  },
+}
+
+const from_email = DEFAULT_MAIL_FROM_ADDRESS
+const from_name = DEFAULT_MAIL_FROM_NAME
+
+const requiredColumns = [
```
lucianak, line 47: Da wären dann am Ende nur die Columns drin, die wir für die Email-Templates brauchen (und im CSV dann auch)

packages/backend-modules/mail/templates/access_recipient_questionnaire_politician.html
```diff
@@ -0,0 +1,125 @@
...
+                    <p>
+                      Sollte dieser Button nicht funktionieren, kopieren Sie bitte den folgenden Link in Ihren Browser: 
+                      <a href="https://tally.so/r/mRd8Bv?uuid={{UUID}}&email={{email}}"
+                        >https://tally.so/r/mRd8Bv?uuid={{UUID}}&email={{email}}</a
+                      >.
+                    </p>
+                    
+                    <p>
+                      Haben Sie noch Fragen? Wir stehen Ihnen gerne unter
+                      <a href="mailto:kontakt@republik.ch"
```
fuenkchen: please use `fragebogen-wahlen@republik.ch` instead of kontakt@

packages/backend-modules/mail/templates/access_recipient_questionnaire_politician.html
```diff
@@ -0,0 +1,125 @@
...
+                          margin: 0px;
+                        "
+                        width="180"
+                        alt=""
+                      />
+                      <br />
+                      Republik AG<br />
+                      Sihlhallenstrasse 1, CH-8004 Zürich<br />
+                      <a href="https://republik.ch">www.republik.ch</a><br />
+                      <a href="mailto:kontakt@republik.ch"
```
fuenkchen: maybe we should use fragebogen-wahlen@republik.ch here too (kontakt@ is the general mail address, I know, but just to be sure, that they use this special email address in case of questions concerning the questionnaire)

packages/backend-modules/mail/script/sendMailsToSegment/politicianQuestionnaireMailing_06-2023.js
```diff
@@ -0,0 +1,191 @@
...
+  'mail:script:politicianQuestionnaireMailing_06-2023',
+)
+
+const argv = yargs
+  .option('dry-run', {
+    default: true,
+  })
+  .option('file', {
+    default:
+      '/home/luciana-republik/plattform/packages/backend-modules/mail/script/sendMailsToSegment/test_politicians.csv',
```
fuenkchen, line 28: please use a relative path here.

fuenkchen (changes requested):
Could you please rename the email templates as well? Atm they have the prefix `access_recipient` which could lead to confusion because they are not access grant related.

Afaik we haven't decided yet if this should be merged into main. If so there should be some documentation (e.g. on how csv file should look like). If not this is not necessary (but could be done as well, in case we want to reuse it again with the code from this PR).

lucianak:
@fuenkchen tbh I don't think we  should resue this code, it's not very generic. I'd say to not merge into main

PR #501, opened 2023-06-21
fix(mdast/remark-preset/span): harden data attribute value parsing

trm217:
CleanShot 2023-06-21 at 10 25 46@2x

(These changes are from a non merged pr in the remark-preset package in the mdast repo. [see #7 in orbiting/mdasst](https://github.com/orbiting/mdast/pull/7))


jstcki (approved):
I don't like what we're doing here (multiple values in _one_ attr) but whatever ¯\_(ツ)_/¯

PR #506, opened 2023-06-28
fix: suggested price comes from correct pledge now

lucianak:
All pledges and membership package options are taken into account for choosing a suggestedPrice at the moment, but only the latest pledge for your own membership should be used. Hence the filter by defaultAmount, which is greater than 0 for the right membership and package option (defaultAmount for gift memberships is 0).


patrickvenetz (approved):
LGTM.

Would add a brief description of issue in Pull Request description field for future code peepz.

fuenkchen (approved):
what @patrickvenetz says

PR #508, opened 2023-06-30
fix(calendar): add weekday limitation attribute

patrickvenetz:
This Pull Request allows to limit available calendar slots to weekdays e.g. Monday to Saturday, but not Sundays.

### DIFF
--- a/packages/backend-modules/calendar/graphql/resolvers/_mutations/bookCalendarSlot.js
+++ b/packages/backend-modules/calendar/graphql/resolvers/_mutations/bookCalendarSlot.js
@@ -1,3 +1,5 @@
+const dayjs = require('dayjs')
+
 const { Roles, transformUser } = require('@orbiting/backend-modules-auth')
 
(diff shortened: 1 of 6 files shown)


packages/backend-modules/calendar/graphql/resolvers/_mutations/bookCalendarSlot.js
```diff
@@ -46,6 +50,11 @@ module.exports = async (_, args, context) => {
     throw new Error(t('api/calendar/slot/error/userBookedAlready'))
   }
 
+  const isOnAllowedWeekday = calendar.limitWeekdays.includes(date.day())
```
lucianak, line 53: limitWeekdays is hardcoded to be Mo-Sa, I guess? How would we handle it if we wanted to have speakers on a Sunday exceptionally? (is that an edge case we want to think of or not?)

lucianak (commented):
LGTM

PR #509, opened 2023-06-30
fix(calendar): improve book, cancel slots

patrickvenetz:
This Pull Requests improves calendar package:

- check when booking or cancelling if slot is in future
- move `CalendarSlot` evaluation into lib/utils
- `CalendarSlot.id` depending on context user


lucianak (commented):
Auch LGTM

PR #510, opened 2023-07-04
fix(auth): check if user is deleted when resolving token

lucianak:
When an access token of any type is resolved, it checks if the user has been deleted (for the cases where the user record has to be kept after deletion for administrative reasons).

### DIFF
--- a/packages/backend-modules/auth/lib/AccessToken.js
+++ b/packages/backend-modules/auth/lib/AccessToken.js
@@ -139,7 +139,7 @@ const resolve = async (token, { pgdb }) => {
   if (userId && scope && expiresAt && hmac) {
     const payload = getPayload({ userId, scope, expiresAt })
     const userRecord = await pgdb.public.users.findOne({ id: userId })
-    if (userRecord.revokedAt) {
+    if (userRecord.deletedAt) {
       return null
     }
     const key = userRecord.accessKey
--- a/packages/backend-modules/auth/lib/AccessToken.js
+++ b/packages/backend-modules/auth/lib/AccessToken.js
@@ -138,20 +138,13 @@ const resolve = async (token, { pgdb }) => {
   const [userId, scope, expiresAt, hmac] = base64u.decode(token).split('/')
   if (userId && scope && expiresAt && hmac) {
     const payload = getPayload({ userId, scope, expiresAt })
-    const userRecord = await pgdb.public.users.findOne({ id: userId })
-    if (userRecord.deletedAt) {
-      return null
-    }
-    const key = userRecord.accessKey
+    const key = await pgdb.public.users.findOneFieldOnly(
+      { id: userId, deletedAt: null },
+      'accessKey',
+    )
     if (key && getHmac(payload, key) === hmac) {
       const scopeConfig = getScopeConfig(scope)
-      return {
-        userId,
-        scope,
-        scopeConfig,
-        expiresAt: moment(expiresAt),
-        hmac,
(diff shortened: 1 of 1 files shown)


lucianak:
Before change: the GraphQl query 
```
{
  user(slug:"0dc984eb-b07e-438d-b4b0-fa8679c7894d") {
    customPackages {
      name
      suggestedTotal
    }
  }
}
```
with a deleted user gives
```
{
  "data": {
    "user": {
      "customPackages": [
        {
          "name": "PROLONG",
          "suggestedTotal": null
        }
      ]
    }
  }
}
```

After change: the same GraphQl query with the deleted user gives
```
{
  "data": {
    "user": null
  },
  "extensions": {
    "tracing": {
      "version": 1,
      "startTime": "2023-07-04T10:12:01.368Z",
      "endTime": "2023-07-04T10:12:01.677Z",
      "duration": 308520362,
      "execution": {
        "resolvers": [
          {
            "path": [
              "user"
            ],
            "parentType": "queries",
            "fieldName": "user",
            "returnType": "User",
            "startOffset": 775412,
            "duration": 307629822
          }
        ]
      }
    }
  }
}
```

packages/backend-modules/auth/lib/AccessToken.js
```diff
@@ -138,13 +138,20 @@ const resolve = async (token, { pgdb }) => {
   const [userId, scope, expiresAt, hmac] = base64u.decode(token).split('/')
   if (userId && scope && expiresAt && hmac) {
     const payload = getPayload({ userId, scope, expiresAt })
-    const key = await pgdb.public.users.findOneFieldOnly(
-      { id: userId },
-      'accessKey',
-    )
+    const userRecord = await pgdb.public.users.findOne({ id: userId })
+    if (userRecord.revokedAt) {
```
patrickvenetz, line 142: `revokedAt` does not exist in user rows.

(This will always default to `true`.)

packages/backend-modules/auth/lib/AccessToken.js
```diff
@@ -139,7 +139,7 @@ const resolve = async (token, { pgdb }) => {
   if (userId && scope && expiresAt && hmac) {
     const payload = getPayload({ userId, scope, expiresAt })
     const userRecord = await pgdb.public.users.findOne({ id: userId })
-    if (userRecord.revokedAt) {
+    if (userRecord.deletedAt) {
```
patrickvenetz, line 142: This will throw an error if `findOne` returns `null`.

```
TypeError: Cannot read properties of null (reading 'deletedAt')
```

patrickvenetz (changes requested):

packages/backend-modules/auth/lib/AccessToken.js
patrickvenetz: Albeit code might work eventually, maybe revert to previous code and extend query like so:

```
   const key = await pgdb.public.users.findOneFieldOnly(
      { id: userId, 'deletedAt !=': null },
      'accessKey',
    )
```

lucianak:
For testing: 

- create token:
```
{
  user(slug: "userId") {
    accessToken(scope: CUSTOM_PLEDGE)
  }
}
```

- check if token resolves to user or to `null`:
```
{
  user(accessToken: "tokenGoesHere") {
    customPackages {
      name
      suggestedTotal
    }
  }
}
```

patrickvenetz (approved):
LGTM

PR #511, opened 2023-07-06
Add Basic CTA

lucianak:
Adds a new type `CallToActionBasicPayload` to the cta API schema, for the most common use case of having text + link + linklabel.

### DIFF
--- a/apps/www/components/CallToActions/useCallToAction.ts
+++ b/apps/www/components/CallToActions/useCallToAction.ts
@@ -11,17 +11,17 @@ import { useCallToActionsQuery } from './graphql/useCallToActionsQuery'
  * }
  */
 export default function useCallToAction(): {
-  data?: CallToAction
+  data?: CallToAction[]
   loading: boolean
   error?: ApolloError
   refetch: () => void
 } {
   const { data, loading, error, refetch } = useCallToActionsQuery()
   const callToActions = data?.me?.callToActions
-  const activeCTA = callToActions?.find((cta) => !cta.acknowledgedAt)
+  const activeCTAs = callToActions?.filter((cta) => !cta.acknowledgedAt)
 
   return {
-    data: activeCTA,
+    data: activeCTAs,
     loading,
     error,
     refetch,
--- a/apps/www/components/CallToActions/useCallToAction.ts
+++ b/apps/www/components/CallToActions/useCallToAction.ts
@@ -1,5 +1,4 @@
 import { ApolloError } from '@apollo/client'
-import { useMemo } from 'react'
 import { CallToAction } from './graphql/CallToAction'
 import { useCallToActionsQuery } from './graphql/useCallToActionsQuery'
 
@@ -9,7 +8,6 @@ import { useCallToActionsQuery } from './graphql/useCallToActionsQuery'
  *  data: CallToAction
  *  loading: boolean
  *  error: ApolloError
- *  handleDismiss: (id: string) => void
  * }
  */
 export default function useCallToAction(): {
--- a/packages/backend-modules/call-to-actions/graphql/resolvers/CallToActionPayload.js
+++ b/packages/backend-modules/call-to-actions/graphql/resolvers/CallToActionPayload.js
@@ -1,6 +1,6 @@
 module.exports = {
-  __resolveType: (cta, args, context) => {
-    if (cta.customComponent) {
+  __resolveType: (payload, args, context) => {
+    if (payload.customComponent) {
       return 'CallToActionComponentPayload'
     }
 
--- a/packages/backend-modules/call-to-actions/graphql/resolvers/CallToActionPayload.js
+++ b/packages/backend-modules/call-to-actions/graphql/resolvers/CallToActionPayload.js
@@ -1,0 +1,9 @@
+module.exports = {
+  __resolveType: (cta, args, context) => {
+    if (cta.customComponent) {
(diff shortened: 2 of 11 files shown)


packages/backend-modules/call-to-actions/graphql/schema-types.js
```diff
@@ -14,11 +14,19 @@ type CallToAction {
   acknowledgedAt: DateTime
 }
 
-type CallToActionPayload {
-  customComponent: CallToActionPayloadCustomComponent!
+union CallToActionPayload = CallToActionComponentPayload | CallToActionBasicPayload
```
patrickvenetz, line 17: This change is likely not enough and is a breaking change.

API resolver for CTA payloads needs to provied `__typename`, returning either `CallToActionComponentPayload` or `CallToActionBasicPayload`. Otherwise, API will throw an error as it can't determine what type one returns.

However, not sure how to determine that.

Furthermore www requires to update queries accordingly:

_before_

```gql
query myCallToActions {
  me {
    id
    callToActions {
      id
      beginAt
      endAt
      acknowledgedAt
      updatedAt
      createdAt
      payload {
        customComponent {
          key
          args
        }
      }
    }
  }
}
```

_after_

```gql
query myCallToActions {
  me {
    id
    callToActions {
      id
      beginAt
      endAt
      acknowledgedAt
      updatedAt
      createdAt
      payload {
        ... on CallToActionBasicPayload {
          text
          linkLabel
          linkHref
        }
        ... on CallToActionComponentPayload {
          customComponent {
            key
            args
          }
        }
      }
    }
  }
}
```

Transition is not smooth. It will lead to errors for some users using old www code accessing new API.

This needs a bit more architecting.

patrickvenetz (changes requested):

packages/backend-modules/call-to-actions/graphql/schema-types.js
patrickvenetz: @jstcki Resolver nicely done.

I assume my comment was misleading. Adding a resolver using `__resolveType` is what resolves `__typename`; not sure why one would want to return an object with `__typename`.

On the breaking change in `query myCallToActions`: That thing is accessed ~30k times a day and I struck me as odd to willfully accept bad requests (until a user is using new www code). But I can accept that as a strategy.

packages/backend-modules/call-to-actions/graphql/resolvers/CallToActionPayload.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  __resolveType: (cta, args, context) => {
+    if (cta.customComponent) {
+      return 'CallToActionComponentPayload'
+    }
+
+    return 'CallToActionBasicPayload'
+  },
+}
```
patrickvenetz, line 9: Handler receives `payload` attribute only, not entire call-to-action row.

```suggestion
module.exports = {
  __resolveType: (payload, args, context) => {
    if (payload.customComponent) {
      return 'CallToActionComponentPayload'
    }

    return 'CallToActionBasicPayload'
  },
}
```

apps/www/components/CallToActions/useCallToAction.ts
```diff
@@ -9,21 +8,20 @@ import { useCallToActionsQuery } from './graphql/useCallToActionsQuery'
...
-  data?: CallToAction
+  data?: CallToAction[]
   loading: boolean
   error?: ApolloError
   refetch: () => void
 } {
   const { data, loading, error, refetch } = useCallToActionsQuery()
   const callToActions = data?.me?.callToActions
-  const activeCTA = callToActions?.find((cta) => !cta.acknowledgedAt)
+  const activeCTAs = callToActions?.filter((cta) => !cta.acknowledgedAt)
```
patrickvenetz, line 21: Just double-checking here:

With this change, you might see many call to actions at once. Thus far intentional idea was to render only one call to action at a time. (If acknowledged, next would rendered and so forth.)

Is that wanted?

packages/backend-modules/call-to-actions/graphql/schema-types.js
jstcki: @patrickvenetz Ah, okay. I just interpreted it like that because it's actually possible to just return objects with `__typename` and then it works. Apparently many roads lead to a resolved gql type 😂 

Yes, I get that … but if we deploy www + api at the same time it should be fine, I think? But after sleeping on it I think we should actually add a new query and deprecate the old one. WDYT?

apps/www/components/CallToActions/useCallToAction.ts
jstcki: Yes/no, it wasn't clear from the code that this was intentional, since we always fetch all of them (even acknowledged ones). So maybe the new and improved query should be changed to always return _one_ active CTA?

patrickvenetz: Sounds to me like a (business) logic question.

Since we've not an actual set of rules, I'd opt for the next best thing: Performance.

It's very likely cheaper to return all CTAs at once than acknowledging one, fetch next and display it?

(Other approach might be to alter API to return next call to action upon acknowledging. Which might feel counter-intuitive: You mutate using a CTA ID but receive a new one or none?)

patrickvenetz: (Settled to not return one active CTA atm.)

PR #512, opened 2023-07-06
Update GraphiQL

jstcki:
Updates GraphiQL to the [latest version](https://github.com/graphql/graphiql)

PS: this uses the packages hosted on unpkg, which will make sure it's always running the latest (major) version.


trm217 (approved):
We removed the WebSocket integration in the updated version.
Perhaps we'll have to add this again in a bit, but for now it's fine.

PR #516, opened 2023-07-10
Set fixedPrice and payMoreSuggestion in DB

lucianak:
Set fixedPrice and payMoreSuggestion in DB (packageOptions) instead of hardcoded lists in the components. 
The code changes that were necessary [here](https://www.notion.so/republikmagazin/Verkauf-Lesha-Buch-2eafb446c22b433b882ac0d46d493edb) would not be necessary anymore in the future, as it would be stored directly in the database if a package should have a fixed price and if the suggestion to pay more should be shown or not.


lucianak:
SQL statements to run for correct data in the new fields:

```
UPDATE "packageOptions" SET "fixedPrice" = false;

UPDATE "packageOptions" 
SET 
"fixedPrice" = true
WHERE "packageId" IN (
  SELECT id FROM packages
  WHERE name IN ('MONTHLY_ABO', 'YEARLY_ABO','LESHA'));
  
UPDATE "packageOptions" SET "payMoreSuggestion" = false;

UPDATE "packageOptions" 
SET 
"payMoreSuggestion" = true
WHERE "packageId" IN (
  SELECT id FROM packages
  WHERE name IN ('DONATE','ABO_GIVE','ABO_GIVE_MONTHS','YEARLY_ABO','LESHA'));
```

packages/backend-modules/republik-crowdfundings/graphql/resolvers/PackageOption.js
```diff
@@ -20,6 +20,12 @@ module.exports = {
 
     return autoPay
   },
+  fixedPrice({ fixedPrice }) {
```
fuenkchen, line 23: Without looking in detail over the whole PR, just a quick comment here:
Not necessary to fetch these two props here. They are already included in the parent data and do not have to be returned again (it's called overfetching).

apps/www/components/Pledge/CustomizePackage.js
```diff
@@ -425,13 +425,8 @@ class CustomizePackage extends Component {
...
     )
-    const payMoreSuggestions = [
-      'DONATE',
-      'ABO_GIVE',
-      'ABO_GIVE_MONTHS',
-      'YEARLY_ABO',
-      'LESHA',
-    ].includes(pkg.name)
+
+    const payMoreSuggestions = pkg.options[0].payMoreSuggestion
```
lucianak, line 429: It would be best to invert this field in the code, so that payMoreSuggestion = true in the DB when the suggestion should be shown, (and not if should *not* be shown as it is now), but I wasn't sure how to do that best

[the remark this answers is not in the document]
jstcki: Do I interpret this correctly: there can be multiple options but in this case we rely on the assumption that there is exactly one? Is there a case where there are zero options? If that's even remotely a possibility, then we'd need to guard against it, either by checking for undefined `pkg.options[0]?.payMoreSuggestion` or checking for any option to have a payMoreSuggestion with `pkg.options.some(opt => opt.payMoreSuggestion)`

jstcki: Yes, I (or we both) can have a look at that

packages/backend-modules/republik-crowdfundings/graphql/schema-types.js
```diff
@@ -94,6 +94,8 @@ type PackageOption {
   minUserPrice: Int!
   userPrice: Boolean!
   autoPay: Boolean
+  fixedPrice: Boolean!
+  payMoreSuggestion: Boolean!
```
jstcki, line 98: I don't see any resolver changes: will these fields automagically show up?

packages/backend-modules/republik-crowdfundings/migrations/sqls/20230710172000-package-option-fixed-price-up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE "public"."packageOptions"
+  ADD COLUMN "payMoreSuggestion" boolean
```
jstcki, line 2: Don't these need a default value of `FALSE`, otherwise GraphQL will complain about returning a `null` value for non-optional fields?

jstcki: Also: do we need to populate these fields for existing packageOptions (we might not if there's a default value)? Shouldn't this be part of the migration?

apps/www/components/Pledge/CustomizePackage.js
jstcki: Wait, the old logic is if the condition in `payMoreSuggestions` is true, then the actual suggestions are `[]`????

packages/backend-modules/republik-crowdfundings/graphql/schema-types.js
lucianak: They do (I tested it), see https://github.com/republik/plattform/pull/516/files/5cd91fc92ab2dcae30603dc0816c389a07237cf5#r1263865068

apps/www/components/Pledge/CustomizePackage.js
lucianak: Yes, it doesn't make a lot of sense to me that it's written like this (hence why I was confused)

PR #517, opened 2023-07-11
Story components x Auf den Punkt

annatraussnig:


packages/styleguide/src/components/ColorLabel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  color: string
+  children: ReactNode
+}) => {
+  const enterEvent = new CustomEvent(COLOR_LABEL_EVENTS.enter, {
+    bubbles: true,
+    detail: { color },
+  })
+  const leaveEvent = new CustomEvent(COLOR_LABEL_EVENTS.leave, {
+    bubbles: true,
+    detail: { color },
```
jstcki, line 23: Note for later: don't just pass color value here (use an ID instead). Color should come from a color palette that also has dark color scheme support.

packages/stories/story-loader/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "identity-obj-proxy": "^3.0.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "tsup": "^6.7.0",
+    "typescript": "^4.7.2",
+    "glamor": "^2.20.40",
+    "d3-scale": "^2.1.2",
+    "d3-format": "^1.4.5",
+    "lodash": "^4.17.21"
+  },
```
jstcki, line 30: I think some extra dependencies snuck in here. Remember to clean them up later.

packages/stories/story-loader/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@republik/story-loader",
+  "version": "0.0.1",
+  "description": "renders a story component based on its name",
+  "private": true,
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.ts",
```
jstcki, line 8: Note for later: use "exports" field instead of main/module/types

packages/styleguide/src/components/ColorLabel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  })
+
+  return (
+    <span
+      {...styles.container}
+      onMouseEnter={() => document.dispatchEvent(enterEvent)}
+      onMouseLeave={() => document.dispatchEvent(leaveEvent)}
```
jstcki, line 30: I am not completely sure if it's a good idea to dispatch custom events on the document (but maybe it's fine). Anyway, we need to think about how this will/should work on mobile.

packages/stories/stories-the-dot-banks/src/toolbox/ScrollyFrame.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [currentStep, setCurrentStep] = useState<number>(1)
+  const [activeColor, setActiveColor] = useState<string>(undefined)
+  const [isFixed, setFixed] = useState<boolean>(false)
+
+  const containerRef = useRef<HTMLDivElement>()
+  const chartRef = useRef<HTMLDivElement>()
+
+  // TODO: some error handling would be good
+  // TODO: refine the scroll/steps logic
+  const handleScroll = () => {
```
jstcki, line 47: This contains a lot of DOM measuring with `getBoundingClientRect`. For performance reasons we should use IntersectionObserver instead.

jstcki (commented):
Left a few notes for future reference. I'm aware that stuff needs to be cleaned up and depends on extraction of theme config/logic into another package, so we won't need to copy it around in different stories.

I haven't looked too closely at template/Publikator stuff 😉 

Noticed an odd thing that turbo now complains that we have too many parallel builds (e.g. when trying to run `yarn dev:www`. Can be fixed by adding `--concurrency` flag but I wonder if there's something to improve (by running fewer build scripts)

But these are all minor details … overall awesome work!

packages/styleguide/src/components/ColorLabel/index.tsx
annatraussnig: yeah the dom events are also a source of concerns for me – for now i just went with the make-it-work approach… if it proves brittle upon testing, i guess we could replace those with some kind of react context.

packages/styleguide/src/components/ColorLabel/index.tsx
annatraussnig: agreed. 🙂 
this refactoring only makes sense once the color stuff is out of the styleguide, though.

packages/styleguide/src/components/ColorLabel/index.tsx
annatraussnig: just tested: on staging and on mobile, seems to be working ok… the app is uncharted territory though.

packages/stories/story-loader/package.json
annatraussnig: some stuff comes from the styleguide bits i ruthelessly copied into the scrolly repo

PR #519, opened 2023-07-19
TS Setup improvements

jstcki:
- Define `moduleResolution` as [`nodenext`](https://www.typescriptlang.org/tsconfig#moduleResolution) (should be `bundler` but that's only supported by Next.js 13+), so the package.json `exports` field is correctly supported (instead of using `main`, `module` etc.).
- Upgrade TypeScript to v5
- Move the TypeScript dependency out of individual packages and into the root of the project (fewer moving parts and dependencies to manage)


lucianak (approved):
LGTM :dog:

PR #521, opened 2023-07-20
Improve Questionnaire Overview Performance

jstcki:
Adds a new query for fetching questionnaire submissions *without* questionnaire metadata.

This improves performance of the questionnaire overview significantly because at the time the submissions are fetched, questionnaire metadata has already been fetched and doesn't have to be re-fetched. Submissions for each question are queried separately which can lead to a lot of concurrent requests (this is still the case and should probably improved at some point). At least this way, each query takes ~1s whereas before it could take 5–10s.


felixmichel (approved):
lgtm

PR #522, opened 2023-07-20
Refactor Theming

jstcki:
Changes how the theming mechanism works. The goal is to get away from our custom React Context based approach towards using CSS custom properties (a.k.a. CSS variables). CSS variables have already been used partially but there's a lot of code which implemented the same things in JavaScript that predates their use. This is neither elegant nor necessary anymore since CSS variables have been supported in all browsers since [at least 2017](https://caniuse.com/css-variables).

The main motivation is to reduce confusion and tediousness around styling elements in a way that respects color scheme preferences. Prior to this PR we had to use our custom `colorScheme` API via React Context to apply a theme value:

```jsx
const styles = css({
  borderRadius: 5
  // Not possible to reference theme values here
});

const SomethingWithThemedColors = () => {
  return <div {...styles} {...colorScheme.set('background', 'primary')} {...colorScheme.set('color', 'text')}>hi</div>
}
```

This is very annoying, since CSS already can do this very well:

```css
.something-with-themed-colors {
  border-radius: 5px;
  background: var(--color-primary); 
  color: var(--color-text);
}
```

… or still using Glamor:

```jsx
const styles = css({
  borderRadius: 5,
  background: 'var(--color-primary)', // in the future we could use a type-safe helper like `token('colors.primary')`
  color: 'var(--color-text)',
});

```

This could technically already be done before but there were some inconsistencies and a lot of complex code which was concerned with deep-merging JS objects and retrieving correct values, which has become completely unnecessary since CSS already does the job much more elegantly.

### Goals

- Remove custom styling implementation in favor of a CSS variable based approach
- Remove legacy code for checking for CSS var support in the browser, it's just not needed anymore
- Replace custom color scheme switching with `next-themes`
- For color scheme overrides (e.g. forcing a color scheme on a particular element) rely only on CSS: it should be as easy as wrapping something in `<div className='dark'>`)
- Not break existing styles

### Non-Goals

- No major API change (ie. `colorScheme.set()` still works as expected, just differently under the hood)
- No theme config refactor and moving it out of the styleguide package, will be done in a subsequent PR
- No big rewrite of code

### Specific things that are still like before
- local color context extension still is a thing (i.e. overriding specific theme values locally). This is still needed in a few places like charts.
- Mapping of color values to theme values (e.g. `#000` → `accentColorMeta`). This should be removed by replacing color values in the DB with variable names but is still currently in use for "formats".
- Some colors switch when dark/light color scheme changes which they shouldn't (e.g. carousel item borders on /audio). This was buggy before already and probably not fixed at a later point.


packages/styleguide/src/components/Colors/ColorContext.tsx
```diff
@@ -24,16 +19,18 @@ const createScheme = (specificColors) => {
...
   const getCSSColor = (color, mappingName = undefined) => {
+    // For backwards compatibility, we map some colors to variable names.
+    // E.g. when mappingName is 'format', '#000' is looked up in mappings.format['#000'] and becomes 'accentColorMeta'
     const mapping = mappings[mappingName] || {}
-    return accessCSSColor(mapping[color] || color)
+
+    return color in mapping
+      ? `var(--color-${mapping[color]})`
+      : color in colorDefinitions
+      ? `var(--color-${color})`
```
trm217, line 32: 🥴

packages/styleguide/src/components/Colors/ColorContext.tsx
```diff
@@ -173,136 +153,52 @@ export const ColorContextLocalExtension: React.FC<{
...
- */
-export const ColorHtmlBodyColors = ({
-  colorSchemeKey = 'auto',
-  colorsObject = DEFAULT_COLORS_OBJECT,
-}: ColorHTMLBodyColorsProps) => {
+export const RootColorVariables = () => {
   return (
     <style
-      key={colorSchemeKey}
+      id='theme-variables'
```
trm217, line 159:
```suggestion
      data-testid='theme-variables'
```

packages/styleguide/src/templates/Front/index.js
```diff
@@ -433,21 +432,19 @@ const createFrontSchema = ({
   const frontTileTeaser = {
     matchMdast: matchTeaserType('frontTile'),
     component: ({ children, attributes, ...props }) => (
-      <ColorContextProvider colorSchemeKey='light'>
```
trm217, line 436: Where is the '.light' class?

packages/styleguide/src/components/Colors/ColorContext.tsx
jstcki: Improved in the [sequel](https://github.com/republik/plattform/pull/528/commits/ffde21a4176e85cfe683015aed2a33a24a6801f7#diff-a7d5b1b2b5d16e2da4dc28c711272ec8a626946f1579e93354156c89f285572fR15-R17)

packages/styleguide/src/templates/Front/index.js
jstcki: It's a) not needed because tile colors are defined via CMS and b) the additional div would mess up the tile layout.

PR #528, opened 2023-08-03
Refactor Theming Part 2

jstcki:
Replaces all direct imports of the `colors` object with CSS vars. Some more simplification of the ColorContext.


apps/admin/components/Tables/utils.js
```diff
@@ -38,22 +37,22 @@ export const tableStyles = {
   selectableRow: css({
     '& td': {
       transition: 'border-color 0.2s',
-      borderTop: `1px solid ${colors.secondary}00`,
-      borderBottom: `1px solid ${colors.secondary}00`,
+      borderTop: `1px solid ${'var(--color-secondary)'}`,
+      borderBottom: `1px solid ${'var(--color-secondary)'}`,
```
jstcki, line 41: @trm217 FYI, this was a weird one, where `00` would be appended to a color string (which obviously doesn't work with a var). No easy way to replicate this behavior but since it's in the admin app I think it's fine to ignore?

PR #533, opened 2023-08-05
fix(calendar): throw error if role is missing

patrickvenetz:
This Pull Request ensures a role, defined in calendars `limitRoles`, is present on signed in user on mutations `bookCalendarSlot`, `cancelCalendarSlot`.

Up until now it did not throw an error (albeit check was present).

Mutations used in https://github.com/republik/vorlesen.

### DIFF
--- a/packages/backend-modules/calendar/graphql/resolvers/_mutations/bookCalendarSlot.js
+++ b/packages/backend-modules/calendar/graphql/resolvers/_mutations/bookCalendarSlot.js
@@ -29,10 +29,12 @@ module.exports = async (_, args, context) => {
     calendar.limitRoles?.length &&
     !Roles.userIsInRoles(user, calendar.limitRoles)
   ) {
-    t.pluralize('api/unauthorized', {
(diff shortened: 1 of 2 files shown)


HakuPiku:
Hi, the original reporter of the bug here. Sorry if it's out of my place to comment here but doesn't this still only check the roles for the `user` object which if `isMyself` is false would be the user loaded based on userId (which is a user-controllable value). 

```
const user = isMyself ? me : await loaders.User.byId.load(userId)
 if (
      calendar.limitRoles?.length &&
      !Roles.userIsInRoles(user, calendar.limitRoles)
    ) {
    ....
    }
```

In short, it checks whether the user for which the calendar is being booked/canceled has the correct role or not. It doesn't, however, ensure that `me` has the correct role. So essentially low priveleged `me` could still book/cancel the calendar for high priveleged `user` with correct roles (given their userId).

lucianak (commented):
LGTM, but I haven't checked for the problem mentioned in the comment

PR #526, opened 2023-08-03
feat: upgrade Next.js to v13

trm217:
Somewhat blocked by #535

PR #535, opened 2023-08-08
refactor(styleguide): (almost) EVERYTHING IS TYPESCRIPT

trm217:
## Reason

Next.js 13 Upgrade PR (#526) is not too happy with certain JS components, as their usage isn't completly correct due to faulty prop definitions (especially when there are no default valeus on optional props).

Anyways, this PR just bulk converts a bunch of component from JS to TS. All done manually, it's fun 🤡


packages/styleguide/src/components/ExpandableLink/ExpandableLink.tsx
```diff
@@ -45,13 +37,21 @@ const Portal: React.FC<{ children: ReactElement }> = ({ children }) => {
     : null
 }
 
+type ExpandableLinkProps = {
+  children?: React.ReactNode
+  attributes: object
```
felixmichel, line 42: could attributes also be optional?

felixmichel (approved):

PR #551, opened 2023-08-18
feat(admin): setup sentry

trm217:


trm217:
TODO: check if zendesk integration still works as expected

PR #556, opened 2023-08-24
Fix Discrete Color Range

### LINKED ISSUE #538: Theming update regression
### Bug Description

The values that the color-context provides no longer seem to be identical after the update, which has caused the chart-editor inside the publikator to break.
In this case, the issue was that before the update, the discrete colors that were resolved in the provided value, were equal to the values defined in theme/colors.js.
But now after the update however, these values no longer end up in the resolved scheme.

### Expected Behaviour

For the colorscheme to be resolved as previously.

### Steps to reproduce

_No response_

### Environment

Other (please specify in version field below)

### Version

_No response_

jstcki:
Discrete colors are now explititly named and the vars referenced in the color range. Also, neutral color is in an array, so we need to pull it out to construct a new range from it.

Fixes #538


trm217 (approved):
LGTM

fire-elmo-gif

PR #561, opened 2023-08-30
feat(publikator): setup sentry

trm217:
- feat(publikator): get rid of express server
- feat(publikator): setup sentry


apps/admin/next.config.js
```diff
@@ -55,6 +55,15 @@ module.exports = withBundleAnalyzer(
...
       return [
+        {
+          source: '/(.*)',
+          headers: [
+            {
+              key: 'Strict-Transport-Security',
+              value: 'max-age=63072000; includeSubDomains; preload',
+            },
+          ],
+        },
```
jstcki, line 66: nit: move these headers to the rule below (source is the same)?

apps/publikator/lib/settings.js
```diff
@@ -1,39 +1,38 @@
...
-export const SG_DYNAMIC_COMPONENT_BASE_URLS = ENV.SG_DYNAMIC_COMPONENT_BASE_URLS
+export const SG_COLORS = process.env.SG_COLORS
+export const SG_FONT_STYLES = process.env.SG_FONT_STYLES
+export const SG_FONT_FACES = process.env.SG_FONT_FACES
+export const SG_LOGO_PATH = process.env.SG_LOGO_PATH
+export const SG_LOGO_VIEWBOX = process.env.SG_LOGO_VIEWBOX
+export const SG_BRAND_MARK_PATH = process.env.SG_BRAND_MARK_PATH
+export const SG_BRAND_MARK_VIEWBOX = process.env.SG_BRAND_MARK_VIEWBOX
+export const SG_DYNAMIC_COMPONENT_BASE_URLS =
+  process.env.SG_DYNAMIC_COMPONENT_BASE_URLS
```
jstcki, line 28: I'm not sure why these SG_ vars even need to be here since they're only used in the SG?

apps/publikator/sentry.client.config.ts
```diff
@@ -0,0 +1,15 @@
+// This file configures the initialization of Sentry on the client.
+// The config you add here will be used whenever a users loads a page in their browser.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs'
+
+Sentry.init({
+  dsn: 'https://064b127cfc1f7794a484500beee38359@o4505721260212224.ingest.sentry.io/4505796048453632',
```
jstcki, line 8: envify this?

apps/publikator/sentry.edge.config.ts
```diff
@@ -0,0 +1,16 @@
+// This file configures the initialization of Sentry for edge features (middleware, edge routes, and so on).
+// The config you add here will be used whenever one of the edge features is loaded.
+// Note that this config is unrelated to the Vercel Edge Runtime and is also required when running locally.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs'
+
+Sentry.init({
+  dsn: 'https://064b127cfc1f7794a484500beee38359@o4505721260212224.ingest.sentry.io/4505796048453632',
```
jstcki, line 9: and this

apps/publikator/sentry.server.config.ts
```diff
@@ -0,0 +1,15 @@
+// This file configures the initialization of Sentry on the server.
+// The config you add here will be used whenever the server handles a request.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs'
+
+Sentry.init({
+  dsn: 'https://064b127cfc1f7794a484500beee38359@o4505721260212224.ingest.sentry.io/4505796048453632',
```
jstcki, line 8: and this

And maybe also the existing ones in admin?

trm217: Will look into this next week

apps/publikator/sentry.edge.config.ts
trm217: Will look into this next week

apps/publikator/sentry.client.config.ts
trm217: Will look into this next week

apps/publikator/lib/settings.js
trm217: Unsure, but that's the way it is across all frontends. If we change it, we should do it across all FEs. Let's do this separately

PR #567, opened 2023-09-05
feat: enable sending articles as emails

trm217:
### Todo

- [ ] Add a nice "MissingNode" component that links to the article page.


trm217:
Will be done in #572

PR #568, opened 2023-09-05
feat: enrich sentry by setting the user on the error

trm217:
### Descriptions

This PR contains the following changes for the admin & publikator apps:
- Ensure the user is being set on the sentry util.
- add custom 404 pages to both apps to allow for easier navigation on 404.

### Commits
- feat(admin): set active-user as sentry user
- feat(publikator): set active-user as sentry user
- feat(admin): add custom 404 page
- feat(publikator): add custom 404 page
- fix(publikator): rename ME_QUERY constant


apps/admin/lib/useMe.js
```diff
@@ -0,0 +1,25 @@
...
+    if (data?.me) {
+      setSentryUser({ id: data?.me.id })
+    } else {
+      setSentryUser(null)
+    }
+
+    return () => {
+      setSentryUser(null)
+    }
+  }, [data?.me])
```
jstcki, line 20: I don't like this as a side effect of useMe. This way, it also doesn't seem to work when the `withMe` HOC is used. So, preferably, the Sentry user should be set explicitly in _app.tsx.

Same concern applies to publikator app.

apps/admin/components/App.js
```diff
@@ -1,22 +1,27 @@
...
+import { useMe } from '../lib/useMe'
 
-const App = ({ router, children }) => (
-  <main>
-    <Head>
-      <title>{router.pathname.replace('/', '')} – Admin</title>
-      <meta name='viewport' content='initial-scale=1.0, width=device-width' />
-      <style>{`
+const App = ({ router, children }) => {
+  useMe()
```
jstcki, line 7: Noted below, it's really not clear why this is used here (I know it's for the side effect of setting the Sentry user but this should be done explicitly)

apps/admin/lib/useMe.js
jstcki: Also, I don't think useEffect should be used? It won't work on the server …

jstcki (changes requested):

jstcki (approved):

trm217:
Will deploy it later once we're done testing the climate-lab tear-down

PR #569, opened 2023-09-07
feat: probelesen email journey

lucianak:
Users who start an access campaign ("Probelesen") will be subscribed to an email journey through syncing their email to a dedicated mailchimp audience. 
Changes:
- added perk to access campaigns to subscribe to mail journey (audience Id to subscribe to is hardcoded but could be a parameter)
- adapted scheduler to archive unsubscribed
- added check in mail.enforceSubscriptions to unsubscribe people from the journey if they buy a membership/subscription
- updated transactionals, follow up transactional will be deactivated in access camapign config in DB


packages/backend-modules/mail/lib/scheduler/archiveUnsubscribed.js
```diff
@@ -9,21 +9,26 @@ module.exports = async (dryRun = false) => {
...
+    MAILCHIMP_PROBELESEN_AUDIENCE_ID,
   ]
 
   audiencesToArchiveUnsubscribed.forEach((audienceId) => {
     archiveUnsubscribed(dryRun, audienceId)
   })
 }
 
-const archiveUnsubscribed = async (dryRun = false, audienceId) => {
+const archiveUnsubscribed = async (dryRun, audienceId) => {
```
jstcki, line 31: a) Why did the dryRun default value change?
b) Maybe this is how it's done elsewhere too but generally I'd prefer an object as function argument with named keys, i.e. `({ dryRun, audienceId }) =>`, because on the caller side you don't need to figure out the order of arguments.

packages/backend-modules/access/lib/perks/subscribeToMailJourney.js
```diff
@@ -0,0 +1,39 @@
...
+const give = async (
+  campaign,
+  grant,
+  recipient,
+  settings,
+  t,
+  pgdb,
+  redis,
+  mail,
+) => {
```
jstcki, line 14: Same comment as below: it seems a bit wild to me that these arguments are not wrapped in an object like
`({
  campaign,
  grant,
  recipient,
  settings,
  t,
  pgdb,
  redis,
  mail,
}) =>`

BUT I wouldn't change it if this is a common pattern.

jstcki (approved):
LGTM, had a few stylistic questions.

packages/backend-modules/mail/lib/scheduler/archiveUnsubscribed.js
lucianak: a) According to sonar it's not good practice to have a default value before a non-default value, and the dryRun parameter with default value is already passed down from the function above
b) good point :)

packages/backend-modules/access/lib/perks/subscribeToMailJourney.js
lucianak: it's a common pattern in the perks lib

PR #570, opened 2023-09-07
feat(www): setup graphql-codegen 🚀

trm217:
### Description

Still got an issue with codegen explained [here](https://github.com/dotansimha/graphql-code-generator/issues/9362). Wait for this to be resolved, before merging.
(Also, we should remove the manual types once this is resolved)

### Todo

- [ ] Wait for the [@graphql/codegen](https://github.com/dotansimha/graphql-code-generator/issues/9362) issue to be resolved
- [ ] Remove manual types in changed files


trm217:
We will add GQL-Codegen for the API with #574 (only to be used in app-dir for the start)

PR #572, opened 2023-09-13
feat: WdWww als Newsletter

lucianak:


apps/www/components/Article/Page.js
```diff
@@ -708,9 +708,8 @@ const ArticlePage = ({
           const format = meta.format
 
           const isFreeNewsletter = !!newsletterMeta && newsletterMeta.free
-          const showNewsletterSignupTop = isFreeNewsletter && !me && isFormat
-          const showNewsletterSignupBottom =
-            isFreeNewsletter && !showNewsletterSignupTop
```
trm217, line 713: Can't understand this change, what the motiviation nor the thoughts behind these changes are. As me is no longer checked, I'm not sure if this still works as expected or not.

apps/www/pages/umfrage/1-minute.js
```diff
@@ -155,7 +155,7 @@ const ThankYou = compose(withT)(({ t }) => {
             <Meta.P key={`p${i}`}>{paragraph}</Meta.P>
           ))}
       </div>
-      <NewsletterSignUp free skipTitle name='ACCOMPLICE' />
+      <NewsletterSignUp free name='ACCOMPLICE' />
```
trm217, line 158: ?

apps/www/components/Auth/NewsletterSignUp.js
```diff
@@ -19,59 +19,69 @@ const SignUp = ({
   t,
   requestSubscription,
   context = 'newsletter',
-  skipTitle,
```
trm217, line 22: Doesn't this break `1-minute.js`?

trm217 (commented):
Can't really say much about this. I have no context of what it should contain and how the final result should look and act.

I am unsure whether the changes in `Article/Page.js` do what they should, due to the huge diff I just can't say…

apps/www/components/Article/Page.js
lucianak: `me` was checked to only show the signup box if you're logged in or not logged in. But it makes more sense to always show the newsletter-signup on a format page for a free newsletter, independent of login state

lucianak: now the logic should be: show the signup on top on the format page, show the signup box on the bottom on the article page, both only for free newsletters

PR #576, opened 2023-09-19
chore: upgrade next.js to v13.5.1

trm217:
Apparently, the production-ready framework has patched 438 bugs 🤔 

https://nextjs.org/blog/next-13-5


jstcki:
Can‘t wait for Next.js Conf where they introduce v14 with 500 new bugs.

jstcki (commented):
LGTM but we need to test drive this obviously. 

Also don’t forget `docs/`.

PR #587, opened 2023-09-20
fix(marketing): guard against missing carousel ref

jstcki:
Fixes an error where the carousel ref could be undefined, possibly because it's used in a delayed callback.


apps/www/components/Marketing/Carousel.tsx
```diff
@@ -112,23 +112,27 @@ const Carousel: React.FC<CarouselProps> = ({
     }
     carousel.addEventListener('scroll', onScroll)
     return () => {
-      carousel.removeEventListener('scroll', onScroll)
+      carousel?.removeEventListener('scroll', onScroll)
```
trm217, line 115: 🤨

trm217 (approved):

PR #591, opened 2023-09-21
chore: setup vscode tasks

jstcki:


.vscode/tasks.json
```diff
@@ -0,0 +1,67 @@
...
+        "group": "dev",
+        "panel": "dedicated",
+        "close": true,
+        "showReuseMessage": false
+      },
+      "isBackground": true,
+      "problemMatcher": []
+    },
+    {
+      "type": "typescript",
```
trm217, line 53: But why?

trm217 (approved):
🤷🏻‍♂️

.vscode/tasks.json
jstcki: It should show all the errors of a project, as opposed to next build which will only show you the first one

PR #602, opened 2023-09-29
chore(api): rename 'DocumentType' gql type to avoid naming conflict with codegen

trm217:


packages/backend-modules/documents/graphql/schema-types.js
```diff
@@ -137,13 +137,13 @@ type Meta {
   sendAsEmail: Boolean
 }
 
-enum DocumentType {
+enum DocumenSchematType {
```
lucianak, line 140:
```suggestion
enum DocumentSchemaType {
```
Typo :) (bei den anderen auch)

trm217:
There are still some issues with the type that I renamed to InternalDocumentNode…

trm217:
From what I see, it is working now. (Currently deployed on ❤️‍🔥-API)

PR #603, opened 2023-10-02
WdWww fixes after launch

lucianak:


jstcki:
Not sure if this PR is still going to be merged but I noticed today that the NL logos need an extra vertical padding _in the image itself_, see https://github.com/republik/plattform/commit/effa9cb4327c1997fb9ffc20cda893956eea40f5

PR #604, opened 2023-10-02
Add leads without active subscription to marketing audience

lucianak:
Add people to marketing audience:
- when they sign up for a free newsletter
- when they sign up for a trial
- when their membership runs out

Update status to not receive marketing mails for people on marketing audience:
- when they get a paid membership

Unsubscribe works through mailchimp, and people who are unsubscribed are archived by the scheduler.

Additional refactoring: treat single and multiple newsletter subscriptions with mail/lib/updateNewsletterSubscriptions instead of two different files


packages/backend-modules/republik/graphql/resolvers/_mutations/updateNewsletterSubscription.js
```diff
@@ -81,15 +81,32 @@ module.exports = async (_, args, context) => {
...
+    const subscriptions = await updateNewsletterSubscriptions(
       {
         user,
+        interests: {},
         name,
         subscribed,
       },
       context,
     )
+    return subscriptions[0]
```
trm217, line 109: Couldn't this yield errors if the returned array is empty?

PR #607, opened 2023-10-04
fix(Questionnaire): remove shallow prop from link because it would lead to 404

jstcki:


trm217 (approved):
ship-mar-gif-by-maria-jose-guzman

PR #610, opened 2023-10-13
feat(www): update ArticleAudioPlayer to render if no audio-source is provided yet, but will be readaloud

trm217:
## Description

This PR enhances the ArticleAudioPlayer component to also be rendered when no audio source is available yet, but the text will be read aloud:

### Screenshot
CleanShot 2023-10-13 at 12 12 34@2x

### DIFF
--- a/apps/www/components/Article/Page.js
+++ b/apps/www/components/Article/Page.js
@@ -829,7 +829,8 @@ const ArticlePage = ({
                                 </div>
                               )}
 
-                              {hasAudioSource && (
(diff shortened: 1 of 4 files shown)


lucianak (commented):
Looks good, did you test what happens when the audio source is added later, if it still works as expected?

trm217:
Not yet.

trm217:
Tested it, found an oversight & fixed it. Seems to work now.

PR #612, opened 2023-10-17
feat(Author): email author variant

ovbm:
Adds new Large Variant to Email Author: 

Screenshot 2023-10-17 at 17 32 01

ToDo:
-[] Test Mailchimp

### DIFF
--- a/apps/publikator/components/editor/modules/author/ui.js
+++ b/apps/publikator/components/editor/modules/author/ui.js
@@ -2,9 +2,10 @@ import { Fragment } from 'react'
 import { createBlockButton, buttonStyles, matchBlock } from '../../utils'
 import injectBlock from '../../utils/injectBlock'
 
-import { Label, Checkbox } from '@project-r/styleguide'
+import { Label, Checkbox, Field } from '@project-r/styleguide'
 import { AuthorSearch } from '@project-r/styleguide/editor'
 import { Text } from 'slate'
+import createOnFieldChange from '../../utils/createOnFieldChange'
 
 export default ({ TYPE, newBlock, editorOptions }) => {
   const InsertButton = createBlockButton({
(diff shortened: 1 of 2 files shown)


apps/publikator/components/editor/modules/author/ui.js
```diff
@@ -33,6 +34,7 @@ export default ({ TYPE, newBlock, editorOptions }) => {
     if (!value.blocks.some(matchBlock(TYPE))) {
       return null
     }
+    const handlerFactory = createOnFieldChange(onChange, value)
```
lucianak, line 37: pedantic but for me `handlerFactory` is not a very clear name, except if it's some kind of standard?

apps/publikator/components/editor/modules/author/ui.js
```diff
@@ -50,6 +52,8 @@ export default ({ TYPE, newBlock, editorOptions }) => {
               portrait: author.value.portrait,
               credentials: author.value.credentials,
             },
+            isLarge: author.value.isLarge,
```
lucianak, line 55: just to be sure I understand correctly: `isLarge` and `greeting` are independent of each other, right? `isLarge` can be false and the component still displays the greeting

lucianak (commented):
Looks good to me, did you check how it looks on mailchimp?

apps/publikator/components/editor/modules/author/ui.js
ovbm: Yes, I found it as a standard way to create Field onChange Events (see links)

apps/publikator/components/editor/modules/author/ui.js
ovbm: they are independent of each other, however, the greeting is only displayed in the "isLarge" case

PR #629, opened 2023-11-01
feat(www): implement signin redirect

trm217:
- feat(www): implement redirect if logged in on '/anmelden'
- feat(www): prevent CTABanner from being rendered in the frame via a prop


apps/www/components/Frame/Header.js
```diff
@@ -190,9 +190,24 @@ const Header = ({
...
-                  if (router.asPath === userButtonLink) {
-                    closeHandler()
+                  if (router.asPath.startsWith(userButtonLink)) {
+                    return closeHandler()
                   } else {
+                    if (userButtonLink === '/anmelden') {
+                      try {
+                        const url = new URL(
+                          userButtonLink,
+                          window.location.origin,
```
jstcki, line 200: Not 100% sure but this will construct a full URL, but when using `router.push`, a relative one is preferable?

So this could be simplified to
```
router.push(`${userButtonLink}?redirect=${encodeURIComponent(router.asPath)}`)
```

apps/www/components/Frame/Header.js
```diff
@@ -190,9 +190,24 @@ const Header = ({
...
+                        )
+                        url.searchParams.set(
+                          'redirect',
+                          encodeURIComponent(router.asPath),
+                        )
+                        return router.push(url)
+                      } catch (e) {
+                        console.error(e)
```
jstcki, line 208: Do we need to log this error? Shouldn't we just `router.push(userButtonLink)` here?

apps/www/pages/anmelden.js
```diff
@@ -11,12 +11,29 @@ import { withDefaultSSR } from '../lib/apollo/helpers'
...
+    if (me && query?.redirect) {
+      try {
+        const redirectTarget = decodeURIComponent(query.redirect)
+        const redirectUrl = new URL(redirectTarget, window.location.origin)
+        // ensure that the redirect target can only be a relative path
+        if (
+          redirectUrl.hostname === window.location.hostname &&
+          redirectTarget.startsWith('/')
+        ) {
+          router.replace(redirectUrl)
```
jstcki, line 28: Shouldn't this be `router.replace(redirectTarget)`?

apps/www/pages/anmelden.js
```diff
@@ -11,12 +11,29 @@ import { withDefaultSSR } from '../lib/apollo/helpers'
...
+        } else {
+          router.replace('/')
+        }
+      } catch (e) {
+        console.error(e)
```
jstcki, line 33: Again, is there no sensible fallback behavior instead of logging the error? Like doing `router.replace('/')` …

jstcki (approved):

PR #630, opened 2023-11-02
feat(www/new-hope): repurpose '/mitteilung' page to allow handling app-sign in for app-dir

trm217:


apps/www/pages/mitteilung.js
```diff
@@ -159,6 +160,8 @@ const Page = ({ router: { query: rawQuery }, t, me, inNativeApp }) => {
 
   const stickyBar = !isProjectR
 
+  const hasRedirect = router.query.redirect
```
jstcki, line 163: Why not just read from the `rawQuery` here? (Or get rid of `withRouter`)

apps/www/src/graphql/republik-api/app-sign-in.ts
```diff
@@ -0,0 +1,60 @@
...
+export type PendingAppSignInResult = {
+  pendingAppSignIn: null | {
+    title: string
+    body: string
+    verificationUrl: string
+    expiresAt: string // ISO date
+  }
+}
+
+export const UNAUTHORIZED_SESSION_QUERY = gql`
```
jstcki, line 22: Do we still need this?

jstcki (approved):
OK, just a few minor details

PR #637, opened 2023-11-08
feat(www): continue implementing share-overlay

trm217:


jstcki (changes requested):
Nice!

- Please center the "Teilen" button the same way as the native one
- Twitter share is broken (on their side): let's remove it?
- Email body is "undefined"

jstcki:
I think it should just be the URL, no need to overcomplicate things and add some generic text that people will delete anyway.

apps/www/src/components/share/share-overlay.tsx
```diff
@@ -2,16 +2,131 @@
...
+      icon: IconLogoTelegram,
+      title: 'Auf Telegram teilen',
+      label: 'Telegram',
+    },
+    {
+      name: 'mail',
+      href: `mailto:?subject=${encodeURIComponent(
+        emailSubject,
+      )}&body=${encodeURIComponent(
+        emailBody + (emailAttachURL ? '\n\n' + url : ''),
```
jstcki, line 107: I think I would just do `encodeURIComponent(url)` here. I don't think emailBody and emailAttachURL are useful.

jstcki (approved):

PR #639, opened 2023-11-08
RIP Wahltindär

jstcki:


ovbm:
Question: Are there elements and files in the styleguide that can be removed due to the removal of the www code?

PR #644, opened 2023-11-15
feat(www): allow dynamically setting the DATO_CMS_ENVIRONMENT over env

trm217:
The value is set to 'main' on www-prod and 'development' on all other www deployements

### DIFF
--- a/apps/www/src/lib/apollo/cms-client.ts
+++ b/apps/www/src/lib/apollo/cms-client.ts
@@ -25,7 +25,7 @@ export function getCMSClient(): NextSSRApolloClient<NormalizedCacheObject> {
         Authorization: `${process.env.DATO_CMS_API_TOKEN}`,
         // forbid invalid content to allow strict type checking
         'X-Exclude-Invalid': 'true',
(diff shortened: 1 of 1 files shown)


apps/www/src/lib/apollo/cms-client.ts
```diff
@@ -25,6 +25,7 @@ export function getCMSClient(): NextSSRApolloClient<NormalizedCacheObject> {
         Authorization: `${process.env.DATO_CMS_API_TOKEN}`,
         // forbid invalid content to allow strict type checking
         'X-Exclude-Invalid': 'true',
+        'X-Environment': process.env.DATO_CMS_ENVIRONMENT,
```
jstcki, line 28: Let's add a check for the DATO_CMS_ENVIRONMENT env var too. If it's `undefined`, the query will fail.

jstcki (changes requested):

apps/www/src/lib/apollo/cms-client.ts
jstcki: Alternatively, just fall back to `'main'`

jstcki (approved):

PR #658, opened 2023-11-29
feat(Goodies): add cookie goodie

ovbm:
ToDo:
Image needs to be replaced with one with a transparent background. Asked Jonas for a new one.


ovbm:
@lucianak ready to deploy.

PR #667, opened 2023-12-07
tooling(www): enable graphql-codegen for republik-api

trm217:
- chore(www): update gql-codegen config
- refactor(api): rename 'DocumentNode' graphql-type to 'ArticleNode'
- feat(www): update graphql codegen setup to correctly generate types for the republik-api and dato-cms


apps/www/codegen.ts
```diff
@@ -46,7 +75,8 @@ const config: CodegenConfig = {
       },
```
jstcki, line 75: Do we need to handle the JSON scalar too?

apps/www/src/app/challenge-accepted/components/teasers/article-teaser.tsx
```diff
@@ -76,7 +74,7 @@ export const ArticleTeaser = async ({ path, image }: ArticleProps) => {
   // and extract the pathname from it.
   const url = new URL(path, process.env.NEXT_PUBLIC_BASE_URL)
 
-  const { data }: { data: ArticleQueryResult } = await getClient().query({
+  const { data } = await getClient().query<ArticleTeaserQuery>({
```
jstcki, line 77: In the `src/graphql/republik-api/article.query` file we should be using the generated `gql` function now instead of the one from `@apollo/client`, so this query ist typed automatically, no?

jstcki: We need to do the same thing in `newsletter-teaser.tsx`

jstcki (commented):
Generally looks good.

1. Shouldn't we use the generated typed queries now (i.e. for all the queries in graphql/republik-api)?
2. I find the naming of the generated directories/files a bit confusing (e.g. stuff like `import { gql } from '@app/graphql/gql/cms/gql'` 😂)

PR #669, opened 2023-12-08
feat(www): re-add fade-in behaviour of page-header in app-dir

trm217:


apps/www/src/components/layout/header/index.tsx
```diff
@@ -57,8 +60,29 @@ const Avatar = ({ me }: { me: MeQuery['me'] }) => {
...
+export function PageHeader({ me }: PageHeaderProps) {
+  const headerRef = useRef<HTMLDivElement>(null)
+  const scrollDirection = useScrollDirection({
+    threshold: 100,
+  })
+  const [elementHeight, setElementHeight] = useState(0)
+
+  useEffect(() => {
+    if (headerRef.current) {
+      setElementHeight(headerRef.current.clientHeight)
```
jstcki, line 83: This will not update when the header height changes?

apps/www/src/components/layout/index.tsx
```diff
@@ -14,16 +15,18 @@ type LayoutProps = {
...
     <div
       className={css({
+        backgroundColor: 'pageBackground',
         minHeight: '100dvh',
         display: 'flex',
         flexDirection: 'column',
       })}
     >
-      <PageHeader />
+      <PageHeader me={me} />
```
jstcki, line 29: This will add a bunch more data to the client which isn't really needed. It would be better to specify which props are actually needed in the PageHeader and only pass those from the server to the client.

apps/www/src/components/layout/header/index.tsx
```diff
@@ -69,12 +80,24 @@ export const PageHeader = async () => {
...
         borderBottomColor: 'divider',
+        position: 'sticky',
+        top: 0,
+        transition: 'transform 0.3s ease-out',
+        zIndex: 100,
       })}
+      style={{
+        transform: `translateY(${
+          scrollDirection === 'down'
+            ? -((headerRef.current?.clientHeight || 0) + 1)
```
jstcki, line 97: I believe this is fine in practice but reading from the DOM in render is generally a no-no?

jstcki (approved):
Overall LGTM 👍 I added 2 comments but they're not blocking IMO.

Maybe remove the generated gql files from this PR?

apps/www/src/components/layout/header/index.tsx
trm217: Don't really have many other options. Either that or the useEffect thing that was there previously.

jstcki: Do we need to know the actual height of the element? We can use something like max header height, which shouldn't make much of a difference even when the actual height is smaller.

jstcki: Just remembered, this can be solved with CSS: we have the value in `var(--sizes-header-height)` (which is responsive)

jstcki: Ah, nevermind, that's without the navigation bar … but it could be solved by adding a totalHeight to sizes.header in the Panda preset

PR #672, opened 2023-12-11
Fixes for Scroll Restoration

jstcki:
This fixes the issue where Next.js' `scrollRestoration` feature (which is generally an improvement) makes things worse with external links. When navigating to an external link and back, the scroll position is lost.

(The cause for this is that `window.history.scrollRestoration` is set to `manual` to allow Next.js to manage the scroll position. This is necessary for internal links but for external links it doesn't work because the current scroll position is not saved when navigating to one.)

This PR doesn't actually solve the root cause (I'm not sure it's even possible – also, within App Router this all works fine). Instead:

1. Each link used in an article is now wrapped in a Next.js Link (this was not the case before!), so internal/relative links can actually use the scrollRestoration feature. 
2. External links are now automatically assigned `target="_blank"`, so in practice the scroll position is not lost there because a new window/tab is opened instead.


trm217 (approved):
LGTM! (I am not 100% sure if I thought about all cases while testing.) 🚢

PR #676, opened 2023-12-14
Event Tweaks

jstcki:
- Fix ordering of upcoming events
- Always format dates/times in the same timezone
- Support for multi-day events
- Support querying CMS items with status other than published (needs to be enabled via env var) and display that status in the event teasers
- Allow exclusion of events from overview via the `unlisted` field


apps/www/src/components/cms/item-status.tsx
```diff
@@ -0,0 +1,23 @@
+import { css } from '@app/styled-system/css'
+
+export const CmsItemStatus = ({ status }: { status: string }) => {
```
trm217, line 3:
```suggestion
export const CMSItemStatus = ({ status }: { status: string }) => {
```

apps/www/src/app/veranstaltungen/components/event-teaser.tsx
```diff
@@ -80,7 +74,7 @@ export const EventTeaser = ({ isPage, isMember, event }: EventProps) => {
               mt: '-0.1lh',
             })}
           >
-            {event.title}
+            {event.title} <CmsItemStatus status={event._status} />
```
trm217, line 77:
```suggestion
            {event.title} <CMSItemStatus status={event._status} />
```

apps/www/src/app/veranstaltungen/components/event-teaser.tsx
```diff
@@ -91,7 +85,9 @@ export const EventTeaser = ({ isPage, isMember, event }: EventProps) => {
               '& a': { textDecoration: 'none' },
             })}
           >
-            <Link href={`/veranstaltungen/${event.slug}`}>{event.title}</Link>
+            <Link href={`/veranstaltungen/${event.slug}`}>
+              {event.title} <CmsItemStatus status={event._status} />
```
trm217, line 89:
```suggestion
              {event.title} <CMSItemStatus status={event._status} />
```

trm217 (approved):

PR #678, opened 2023-12-20
feat(www): improve gql codegen

trm217:


jstcki (approved):

apps/www/graphql/configs/codegen-dev.ts
```diff
@@ -0,0 +1,51 @@
...
+          Date: 'string',
+          DateTime: 'string',
+        },
+      },
+      plugins: [],
+    },
+  },
+  // Ensure that the generated files are formatted directly to not cause diffs if there are no changes
+  hooks: {
+    afterOneFileWrite: ['prettier --write'],
```
jstcki, line 47: Not a new change but we could also just prettier-ignore the generated files? Better for performance?

trm217: Added it to .eslintignore + .prettierignore

PR #680, opened 2023-12-22
UI Package Setup

jstcki:


jstcki:
As discussed with @trm217, we'll abandon this route and not add yet another package. The monorepo makes everything much more complicated than it needs to be, with twice the configs, and tight coupling between the packages anyway. The better strategy is to try to consolidate the frontend apps into one, and build components there, with a single config.

PR #685, opened 2024-01-05
feat(www): implement lightswitch in footer (app-dir & pages-dir)

trm217:
## Description

Until now the theme-toggle was hidden in the user-nav, which also made it inaccessible for non logged in people.
Now there is a second theme-toggle located in the footer of every page. 💡
CleanShot 2024-01-05 at 16 57 15@2x
CleanShot 2024-01-05 at 16 58 45@2x

### DIFF
--- a/apps/www/src/app/challenge-accepted/page.tsx
+++ b/apps/www/src/app/challenge-accepted/page.tsx
@@ -51,7 +51,9 @@ export async function generateMetadata(
 }
 
 export default async function Page({ searchParams }) {
-  const { data } = await getCMSClient().query({
+  const {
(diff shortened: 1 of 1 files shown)


ovbm:
Love it. LGTM, merge and ship.

apps/www/src/components/lightswitch.tsx
```diff
@@ -0,0 +1,75 @@
+'use client'
+
+import { css } from '@app/styled-system/css'
```
ovbm, line 3: So we can use the new css function in the www directory?

ovbm (approved):

PR #689, opened 2024-01-16
fix(Profile): show prolitteris id for all users with at least one doc…

ovbm:
- Shows Prolitteris ID Field if a user has at least one item in `user.documents`
- Also fixes the textarea for pgp apparently having an invalid prop `row` instead of `rows`


lucianak (approved):
looks good 👍

lucianak (approved):

PR #692, opened 2024-01-16
Goto-App

patrickvenetz:
This Pull Requests adds new app: "goto".

It will forward a visitor to https://www.republik.ch, preserving path and query parameters. Say, you click on https://goto.republik.ch/angebote?package=FOOBAR. App will forward you to https://www.republik.ch/angebote?package=FOOBAR.

It helps us to leave app when clicking on a republik.ch link; apart from exceptions, all republik.ch links are Universal Links and will open in Republik-App. If link is "goto-ified" however, it will open in browser and forward in-browser to https://www.republik.ch.

Changes include expanding docs with goto-app and information about linking within Republik empire.


jstcki:
I'm skeptical of setting up another Next.js app for this. Seems like an awful lot of code and new dependencies (Tailwind??) we add to our codebase for something as trivial as a redirect, which can be done with a single static HTML file? 🤔

patrickvenetz:
@jstcki You're right, it was an awful lot of code. Mainly due to apps original authorization feature, which we dropped along the way.

Did refactor accordingly.

Would like to keep Next.JS app, as it allows to serve an path and forward a user in a more user friendly manner e.g. `goto.republik.ch/foo/bar?fizz=buzz` to `www.republik.ch/foo/bar?fizz=buzz`.

(Assuming a static HTML file would require target path in a parameter e.g. `goto.republik.ch/index.html?url=%2Ffoo%2Fbar%3Ffizz%3Dbuzz` to `www.republik.ch/foo/bar?fizz=buzz`?)

And on the account of dependencies:Apart from Tailwind CSS, there are no new dependencies. @trm217 helped out with UI (🙇) and introduced Tailwind CSS hereafter. Since I've no objection to either keep or remove it, maybe settle this internally?

jstcki:
I created another app to demo what I meant: https://github.com/republik/goto (I haven't yet added the link to manually click if the redirect doesn't work).

Pathname-preserving redirects work fine, e.g. https://goto-republik.vercel.app/angebote?package=ABO

So my proposal would be to scrap the Next.js app, keep the docs from this PR and link the proper Vercel app to above-mentioned repo.

PR #694, opened 2024-01-17
New backend module for referral campaigns

lucianak:
New backend module for referral campaigns, for 2024 referral campaign and hopefully also future referral campaigns.

Contains:
- new module referral-campaigns
- referral campaign tables
- graqhql resolvers
- ...


jstcki:
> This is how you can query the relevant information for the campaign

Nice. And how do I query a user by alias? Can I just pass it to the slug argument (e.g. `{ user(slug: "xyz") { id } }`)? I didn't spot any resolver change related to this …

packages/backend-modules/auth/graphql/schema-types.js
```diff
@@ -47,6 +47,7 @@ type User {
   # true: consented
   # false: consent revoked
   hasConsentedTo(name: String!): Boolean
+  alias: String!
```
lucianak, line 50: Should the alias rather be in the referral-campaign package? Or do you think it's a generic alias that we might also use for other features?

lucianak, line 50: I think the alias should not be reuqired but be optional here

packages/backend-modules/auth/lib/alias.js
```diff
@@ -0,0 +1,83 @@
...
+    if (attempts < HASH_LENGTH_ATTEMPTS) {
+      attempts++
+    } else {
+      length++
+      attempts = 0
+    }
+    totalAttempts++
+
+    if (totalAttempts > MAX_ATTMEPTS) {
+      throw new Error('Could not generate alias')
```
lucianak, line 74: I think I would not throw here but return null, so that we don't break the User resolver if the alias cannot be generated.

packages/backend-modules/auth/graphql/resolvers/User.js
```diff
@@ -83,4 +84,7 @@ module.exports = {
       })
     }
   },
+  alias: async (user, _, { pgdb }) => {
+    return user.alias || (await generateUserAlias(user, pgdb))
```
lucianak, line 88: I would only return the user alias for your own user and for admin/supporter roles. Maybe something like
```suggestion
    if (Roles.userIsMeOrInRoles(user, me, DEFAULT_USER_ACCESS_ROLES)) {
        return user.alias || (await generateUserAlias(user, pgdb))
    }
```

packages/backend-modules/referral-campaigns/lib/referral-handler.js
```diff
@@ -0,0 +1,99 @@
...
+    pgdb,
+  )
+
+  if (!referrerId) {
+    debug('referral-handler')('no referrer found for pledge', pledge?.id)
+    return
+  }
+
+  const campaign = await pgdb.public.campaigns.findOne({
+    id: payload?.utm_campaign,
```
lucianak, line 78: Maybe we shouldn't force using utm_campaign for the campaign-id, because we also use utm_campaign parameters for matomo tracking etc. I would suggest something like ref_campaign or ref_campaign_id, something that is not otherwise used

packages/backend-modules/referral-campaigns/migrations/sqls/20240117120959-referral-campaign-ddl-down.sql
```diff
@@ -0,0 +1,5 @@
+-- migrate down here: DROP TABLE...
+
+DROP TABLE refferals;
```
lucianak, line 3:
```suggestion
DROP TABLE referrals;
```

packages/backend-modules/republik-crowdfundings/lib/Mail.js
```diff
@@ -1307,4 +1307,79 @@ mail.getPledgeMergeVars = async (
   ]
 }
 
+mail.sendFutureCampaignSenderRewardMail = async (
```
lucianak, line 1310: Should probably be renamed and generalised (note for myself for later)

packages/backend-modules/republik-crowdfundings/lib/futureCampaignSenderReward.js
```diff
@@ -0,0 +1,132 @@
+const debug = require('debug')('crowdfundings:lib:futureCampaignSenderReward')
```
lucianak, line 1: this code should be removed before deploying, is now replaced by handleReferral

packages/backend-modules/republik-crowdfundings/lib/payments/Pledge.js
```diff
@@ -74,6 +79,8 @@ const afterChange = async ({ pledge }, context) => {
     pledge.status === 'SUCCESSFUL' && refreshPotForPledgeId(pledge.id, context),
     pledge.status === 'PAID_INVESTIGATE' &&
       slack.publishPledge(user, pledge, 'PAID_INVESTIGATE'),
+    handleReferral(pledge, { pgdb: context.pgdb }),
```
lucianak, line 82: only if pledge was successful?

packages/backend-modules/translate/translations.json
```diff
@@ -740,6 +740,26 @@
       "key": "api/email/membership_deactivated/subject",
       "value": "Ihre Mitgliedschaft wurde deaktiviert"
     },
```
lucianak, line 742: Also to be removed/adapted to generic referral campaigns

lucianak (commented):
Looks good, thanks! I added a few comments, I didn't test it yet though

packages/backend-modules/auth/graphql/schema-types.js
jstcki: If generic, should it rather be named something like `shortId` which aligns more with our intent? `alias` sounds more like something users can choose for themselves.

packages/backend-modules/auth/lib/alias.js
jstcki: Also a reason to make the field optional (even if the resolver errors, only this field will be null in the gql response, not the whole User)

jstcki:
Superseded by #703

PR #703, opened 2024-01-19
feat(api): referral-campaign backend

trm217:
## Todo

- [x] `campaigns` resolver
- [x] `campaign` resolver
- [x] define what happens when a referral is not for an active campaign
- [x] handle campaign and user ref with different query-params than `utm_*`
- [x] in handleReferral trigger check if an award should be granted. (alternatively do that in a cron-job (i think this might be the better option)
- [x] add migrations for missing tables (see schema in docs/pages/software/applications/api/campaign.mdx)
- [x] implement logic to grant awards
- [x] add parameter to user query to resolve user using the referralCode
- [x] tests (???)
- [x] ~~caching?~~
- [x] move referralCode stuff from auth module to referral-campaign module
- [x] update transactional texts
- [x] update merge user and delete user with new tables etc
- [x] new/updated variables for transactionals:
   - [x] number of referrals
   - [x] bonus: name of referred person for sharer
   - [ ] bonus: ist erstes Abo bzw wird nicht für die onboarding journey angemeldet
   - [ ] bonus: aktueller Stand der Zielerreichung
- [ ] bonus: prevent buying YEARLY_ABO in backend if person already has ABO or BENEFACTOR_ABO


packages/backend-modules/types/index.d.ts
```diff
@@ -31,6 +31,7 @@ export interface UserRow {
   portraitUrl: string
   roles: string[]
   email: string
+  alias: string | null
```
jstcki, line 34: this should be `referralCode`

hdahlheim (approved):

PR #707, opened 2024-01-26
feat: add autoPayMutable to membershipType in DB

lucianak:
autoPayIsMutable is now resolved from membership active and renew-status and the autoPayMutable field, instead of relying on hardcoded logic


packages/backend-modules/republik-crowdfundings/migrations/sqls/20240126153918-add-autopay-mutable-field-up.sql
```diff
@@ -0,0 +1,3 @@
+-- migrate up here: CREATE TABLE...
+ALTER TABLE "public"."membershipTypes" ADD COLUMN "autoPayMutable" boolean NOT NULL DEFAULT false;
+UPDATE "public"."membershipTypes" SET "autoPayMutable" = TRUE WHERE "name" = 'ABO';
```
patrickvenetz, line 3: Looks like `BENEFACTOR_ABO` is missing?

PR #708, opened 2024-01-31
Introduce MONTHLY_ABO_AUTOPAY

patrickvenetz:
This Pull Request enables us to sell (monthly) subscriptions using AutoPay feature. Creating new Stripe subscriptions is disabled.


hdahlheim:
As mentioned in #710, due to contracts not working out, we won't be continuing with this change. Thanks for the effort put into the work.

PR #713, opened 2024-02-07
feat(auth): add author role

hdahlheim:
Authors should be able to have a public profile even if they don't have an active subscription.


ovbm (approved):
LGTM

PR #715, opened 2024-02-09
Fix Publikator Crash for Articles with Embedded Comments

jstcki:
This would cause a crash because parent is a readonly object. Instead we mutate the children in place.


hdahlheim:
LGTM 👍

hdahlheim (approved):

jstcki:
(Most lines changed by prettier 🤦)

PR #720, opened 2024-02-19
feat(backend): add backend module generator script

hdahlheim:
Add a script for generating a typescript backend module. 

The script can be invoked by running `yarn gen:bm`


lucianak (commented):
LGTM 🙏 

I would also add a readme to the gen-backend-module to explain what it does, and add a readme file to generated modules by default :)

PR #723, opened 2024-02-23
feat(api): audio-source metadata enriched with file byteSize

trm217:
In addition to the durationMS being fetched for an attached audioSource, the byteSize is now also attached to the commit meta-data.

### DIFF
--- a/packages/backend-modules/publikator/script/migrateAudioSources.ts
+++ b/packages/backend-modules/publikator/script/migrateAudioSources.ts
@@ -12,7 +12,9 @@ const {
 } = require('@orbiting/backend-modules-base')
 
 const { maybeDelcareMilestonePublished } = require('../lib/postgres')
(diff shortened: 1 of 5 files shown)


hdahlheim (approved):

packages/backend-modules/publikator/script/migrateAudioSources.ts
```diff
@@ -102,7 +104,7 @@ const handleBatch = async (rows: any[], count: number, pgdb: any) => {
       return
     }
```
lucianak, line 106: This script probably woudln't work correctly like this, you would also need to check for audioSourceByteSize in the checks above. But afaik this script was used to add a synthetic voice to all past articles, so probably it isn't that relevant

lucianak (commented):
Looks good

PR #724, opened 2024-02-27
maintenace: sentry in admin & publikator

trm217:
Bump Sentry dependency & add custom error pages to both apps and forward the error to sentry


jstcki (approved):
Looks magic but okay 😁

PR #726, opened 2024-02-27
Yearly Abo Additional Winbacks

ovbm:


hdahlheim (approved):

ovbm:
Question: Should we rename membership_owner_prolong_yearly_abo_winback_14.html to membership_owner_prolong_yearly_abo_winback_16.html, since we changed the time to +16 (from +14).

PR #731, opened 2024-02-29
refactor(mdast-react-render): separate email rendering export

jstcki:
In order to use the other exports of this package in server components, the email rendering parts can't be exported in the same file because of the dependency on `react-dom/server` (importing it will cause an error in RSC).


lucianak (approved):
LGTM, maybe test to publish a newsletter or wdwww on love and see if it still renders on mailchimp (I don't see why it wouldn't though)

jstcki:
Tested ✅

PR #732, opened 2024-02-29
feat(prolitteris): ProLitteris backend module

hdahlheim:
- [x] ProLitteris backend module conforming to api spec
- [x] Batch submission script


hdahlheim:
Because ProLitteris entries cannot be changed after the first submission, I'm hesitant to integrate the article submission directly into Publikator.

Unless we improve the author attribution inside Publikator, this could result in wrong or incomplete author information being submitted to ProLitteris.

PR #743, opened 2024-03-07
Bugfix for problem with activating new memberships in grace period

lucianak:
The bug: when a new membership is bought during the grace period, the old membership is cancelled immediately because the old membership is overdue, but the new membership is not activated because the old membership is still active. 

Proposed fix: the old membership is not cancelled immediately, the new membership is added as dormant, and the changeover scheduler handles the transition


lucianak:
Changed behaviour not while the new membership is created, but how the old membership is cancelled and the changeover is handled. It works for the yearly abo changeover, but we need to make sure it works for other changeovers as expected.

PR #747, opened 2024-03-12
feat(segments): segment with referral-link for 2024 campaign

lucianak:


hdahlheim (approved):

hdahlheim:
@lucianak should we merge this or close the pr because it was a one of change? If the plan is to merge then we should fix the formatting bug.

PR #774, opened 2024-04-02
feat: improved mailchimp lead tracking

hdahlheim:
Formalises part of a workflow to track conversions of newsletter subscribes that got added to mail chimp by zapier.

## Reason for this change 
 Currently this workflow uses a temporary DB table which is not tracked by any migration.

## Changes
- Add new DB table `leadTracking`
- Add import script for mail chimp csv


hdahlheim:
Import script works as intended on staging ✅

packages/backend-modules/lead-tracking/README.md
```diff
@@ -0,0 +1,16 @@
+# How to use the import script
+
+To import emails from mail chimp csv use the following command
```
lucianak, line 3:
```suggestion
To import emails from an exported mail chimp csv-file use the following command
```
Very nitpicky, my first thought upon reading was it would directly import from mailchimp so maybe something like this would be more explicit 🙈

lucianak (approved):
LGTM 👍

PR #775, opened 2024-04-03
fix(publikator): add uuid as a dependency

trm217:
I stumbled upon [this error](https://republik-ag.sentry.io/issues/5110815261/?project=4505796048453632&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=2) and found [this](
https://github.com/republik/plattform/blob/63395e7d53c52ebb6e07e23bd092cd10847c6859/apps/publikator/components/Publication/useValidation.tsx#L13).

### DIFF
--- a/apps/publikator/package.json
+++ b/apps/publikator/package.json
@@ -117,6 +117,7 @@
     "unified": "^7.0.0",
     "unist-util-visit": "^2.0.3",
     "url": "^0.11.0",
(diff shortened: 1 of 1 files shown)


jstcki (approved):
OK, please make sure this actually works, since we have multiple uuid version flying around IIRC

PR #779, opened 2024-04-03
Small fixes for campaign reward scheduler

lucianak:
Small fixes for campaign rewards scheduler to improve efficiency for future campaigns. 
- decrease scheduler frequency, now runs every 6h
- ignore non-personal referrals


lucianak:
After checking with support, manually added entries to userCampaignRewards so that possible rewards for referrers without current active membership are not checked by the scheduler every time it runs.

PR #801, opened 2024-04-17
feat: require adding report-reason when reporting comments

trm217:
- **feat(api): allow adding a description when reporting a comment**
- **feat(web): allow adding text to comment report**


hdahlheim (approved):
Backend looks good to me.

jstcki (commented):
Looks good, let's put this on staging and let support check if the wording is okay

PR #808, opened 2024-04-22
fix(www): ensure ProLitteris API func is able to parse an IP from a comma separated list

trm217:
## Description

Attempt to solve [this issue](https://republik.sentry.io/issues/23618/?project=4507101768908880&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0), where a string with comma separated IPs is not being handled correctly and yields an error in the prolitteris API route.

## Commits

- Extract logic to get a single IP from the request into a dedicated function.
- Handle case multiple IPs being supplied in a comma separated string.

### DIFF
--- a/apps/www/pages/api/prolitteris.ts
+++ b/apps/www/pages/api/prolitteris.ts
@@ -31,27 +31,49 @@ function getHash(input: string | number | object): string {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ovbm (approved):
LGTM, unit test for getRequestIP?

PR #814, opened 2024-04-26
feat(www): setup 'apollo-link-sentry'

trm217:


jstcki (approved):
Don't forget to remove the hand-rolled SentryErrorLink (now or later)

PR #818, opened 2024-04-26
feat(container/api-test): testcontainer setup for api

hdahlheim:
The scope of this PR evolved a bit

- Adds a cloud native Buildpack build setup to allow building the app containers as close to production code as possible
- updates the migration setup to allow programatic db creation; migration and seeding
- adds a example test to the api app `boot.test.js` to show how to start the test container setup


apps/api/__test__/boot.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+  test('api is responding', async () => {
+    const res = await fetch('http://localhost:5010/graphiql')
+
+    await expect(res.status).toBe(200)
+  })
+
+  afterAll(async () => {
+    await server.close()
+    composeEnv.down()
```
jstcki, line 43: Doesn't this need to be `await`ed?

packages/backend-modules/migrations/script/db/create.js
```diff
@@ -1,80 +1,17 @@
...
   process.exit(1)
 }
 
-PgDb.connect({
-  application_name: 'setup',
-  connectionString: stripDatabase(DATABASE_URL),
-})
-  .catch((e) => {
-    console.error('Unable to connect to database. Error:', e.message)
-    console.log('This is likely due to a faulty DATABASE_URL.')
```
lucianak, line 59: Could also be that Docker isn't running 😅

apps/api/__test__/boot.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    await seedSampleData(process.env.DATABASE_URL)
+
+    server = await require('../server').start()
+  }, 120000)
+
+  test('api is responding', async () => {
+    const res = await fetch('http://localhost:5010/graphiql')
+
+    await expect(res.status).toBe(200)
```
lucianak, line 38:
```suggestion
   expect(res.status).toBe(200)
```
Is await necessary here?

packages/backend-modules/migrations/index.js
```diff
@@ -0,0 +1,15 @@
+const {
+  createDB,
+  migrateUp,
+  migrateDown,
+  migrateRest,
```
lucianak, line 5:
```suggestion
  migrateReset,
```

packages/backend-modules/migrations/index.js
```diff
@@ -0,0 +1,15 @@
...
+  seedSampleData,
+} = require('./lib')
+
+module.exports = {
+  createDB,
+  migrateUp,
+  migrateDown,
+  migrateRest,
```
lucianak, line 13:
```suggestion
  migrateReset,
```

docker-compose-test.yml
```diff
@@ -0,0 +1,77 @@
...
+
+  kibana:
+    image: docker.elastic.co/kibana/kibana:7.16.2
+    ports:
+      - 5601:5601
+    depends_on:
+      - elastic
+
+  redis:
+    image: redis:4.0.14-alpine
```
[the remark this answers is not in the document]
hdahlheim: Copied from the dev docker compose, will fix that.

packages/backend-modules/maillog/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@orbiting/backend-modules-maillog",
+  "version": "0.0.1",
+  "private": true,
+  "description": "maillog-migrations",
+  "main": "index.js",
```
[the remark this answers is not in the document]
hdahlheim: I will remove the main script. I just created a placeholder package so that the files are included in the api build.

packages/backend-modules/migrations/script/db/create.js
```diff
@@ -1,80 +1,27 @@
...
+    const dbName = await createDB(
+      DATABASE_URL,
+      process.env.NODE_ENV !== 'production',
     )
-
-    if (envData !== updatedEnvData) {
-      await fs.writeFile(path, updatedEnvData, 'utf8')
-      console.log(`DATABASE_URL in "${path}" updated.`)
-    }
+    console.error(`Database "${dbName}" created.`)
```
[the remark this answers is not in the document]
hdahlheim: `console.error` just causes the message to be logged to `stderr` instead of `stdout`. In other environments, it is common practice to print debug info to `stderr` instead of `stdout` so that `stdout` can be used for piping to different programs.

Because this message has no programmatic use and is only for human debug purposes, I decided printing to `stderr` is the right choice.

apps/api/__test__/boot.test.js
hdahlheim: your right the `compseEnv.down()` should also be awaited.

package.json
```diff
@@ -55,6 +56,7 @@
     "migrate:create": "packages/backend-modules/migrations/script/create.js",
     "migrate:up": "DOTENV_CONFIG_PATH=apps/api/.env NODE_OPTIONS='-r dotenv/config' db-migrate up --migrations-dir packages/backend-modules/migrations/migrations",
     "migrate:down": "DOTENV_CONFIG_PATH=apps/api/.env NODE_OPTIONS='-r dotenv/config' db-migrate down --migrations-dir packages/backend-modules/migrations/migrations --verbose",
+    "heroku:migrate:up": "db-migrate up --migrations-dir packages/backend-modules/migrations/migrations",
```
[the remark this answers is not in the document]
hdahlheim: True, but it has been like this in the other migration commands. This might become a problem when we switch to pnpm eventually.

We could wrap db-migrate in a script inside the migrations package and only reference this script, but then again, the migrations package is also not explicitly listed as a workspace dependency, so we cannot be sure that the migrations package is actually there.

packages/container-builder/index.js
```diff
@@ -0,0 +1,151 @@
...
+
+function getOSName() {
+  switch (process.platform) {
+    case 'darwin':
+      return 'macos'
+    // support for other operating systems can be added later
+    default:
+      throw new Error('unsupported os')
+  }
+}
```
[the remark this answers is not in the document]
hdahlheim: I don't seed the need to add a lib for supporting other os strings. Currently we only need to support macOS and in the feature we should migrate to proper docker containers.

PR #819, opened 2024-05-02
feat(determine author gender): get rid of the gsheets

hdahlheim:


hdahlheim:
There are plans to solve this differently

PR #830, opened 2024-05-08
chore(frontends): bump jest + @testing-library/react + @testing-library/jest-dom to latest

trm217:


jstcki (approved):
If it works, it works!

PR #833, opened 2024-05-10
Remove References to Translation GSheets

jstcki:
We decided it's less confusing to manage translations directly in code 😏


hdahlheim (approved):
Nice one!

PR #836, opened 2024-05-14
fix(assets/screenshot): only store screenshot if status code < 300

trm217:


hdahlheim (approved):
Simple, reasonable change 👍

PR #839, opened 2024-05-17
Remove S3 Caching from Assets Render

jstcki:
Removes the `permanentCacheKey` logic from both the Frontend (it's only used on the overview) and Asset Server. Caching images on S3 is redundant since we're using a CDN already. Also optimizes overview image loading by requesting smaller sizes and specifying `format=auto` which will favor WebP over PNG.


hdahlheim (approved):
78 Lines less to maintain. Love to see it!

PR #840, opened 2024-05-17
chore(styleguide): use 'tsup' instead of 'tsc' to compile styleguide

trm217:
This PR bundles the styleguide with `tsup` rather than `tsc`.
We bundle the code both as CJS and ESM to allow using it in our frontends, as well as the API.
Since the exports in key in the styleguides package.json now points to the bundled ESM rather than the `ts` files within the src directly (which was wrong to begin with 🙄), running styleguide + <any of our frontends> in parallel in dev-mode might feel less responsive (not that it was fast up until now 😛)


packages/styleguide/tsup.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig } from 'tsup'
+
+const config = defineConfig({
+  entryPoints: ['src/lib.ts', 'src/editor.ts', 'src/chart.ts'],
+  format: ['cjs', 'esm'],
+  dts: true,
+  sourcemap: true,
+  clean: true,
+  minify: true,
```
jstcki, line 9: Is minify necessary? Might be faster if we don't (at least in dev)

jstcki:
Another thing I noticed: editor integration is broken now, Intellisense will point to the minified chunk file in dist/

packages/styleguide/tsup.config.ts
```diff
@@ -0,0 +1,18 @@
...
+  target: 'es6',
+  external: ['react', 'react-dom'],
```
jstcki, line 11: This should not be necessary. `dependencies` and `peerDependencies` are automatically excluded.

jstcki:
Intellisense is working now but I get a new error during development, whenever I change something in styleguide:

image

This shows up during Fast Refresh but goes away after a page reload (also when Next.js reloads on its own)

jstcki:
Also this error now shows up in the console:

image

I think is because the `.js` loader should be different in dev than prod

jstcki:
For the minify option, [this](https://tsup.egoist.dev/#conditional-config) looks like a nicer approach than checking for NODE_ENV:

```
export default defineConfig((options) => {
  return {
    minify: !options.watch,
  }
})
```

jstcki:
> Intellisense is working now but I get a new error during development, whenever I change something in styleguide:
> 
> image
> This shows up during Fast Refresh but goes away after a page reload (also when Next.js reloads on its own)

Seems to be better when `clean` is turned off (again, make this dependent on `options.watch`?)

jstcki (approved):

PR #848, opened 2024-05-27
Update to node.js 20

hdahlheim:
Tested locally and on love. No issues found so far.


apps/admin/package.json
lucianak, line 1: Do we also need to change line 75 `"@types/node": "18",` to 20? Same in docs/package.json, www/package.json, nextjs-apollo-client/package.json, styleguide/package.json

lucianak (commented):
Do we need to update here too? It seems deprecated: plattform/.github/workflows_archive/styleguide.yml
What about the "@types/node": "18" entries in the package.json files (see comment)?
That's all I found, seems to run well on my machine too with node 20 👍

PR #849, opened 2024-05-27
Native Image Lazy-Loading

jstcki:
The [`loading="lazy"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#loading) attribute is supported natively on img elements across all browsers since 2022. This removes the custom JS-based code for doing exactly that.

In addition to reducing the JS footprint and improving the loading behavior in general, this makes it easier to load images when rendering PDFs using Playwright.


jstcki (commented):
Note: this doesn't remove the LazyLoad component completely, since it's still used on the Marketing/Overview pages.

PR #852, opened 2024-05-29
feat(api): resolve CORS dynamically

trm217:


hdahlheim:
I reviewed the code, and although this version works, I would like to propose using version #854 because it is a bit more straightforward.

PR #862, opened 2024-06-07
Bye Bye, SG_ Environment Vars

jstcki:


hdahlheim (approved):
Wonderfull change! 

I'm not 100% sure if putting the logos into a JSON is the best way, but it works for me.

PR #864, opened 2024-06-11
feat(backend): add pg-boss based worker queue

hdahlheim:


packages/backend-modules/job-queue/lib/queue.ts
```diff
@@ -0,0 +1,95 @@
...
+  async clearStorage() {
+    return this.pgBoss.clearStorage()
+  }
+
+  async startWorkers() {
+    const workers: Promise<string>[] = []
+
+    for (const worker of this.workers.values()) {
+      workers.push(this.pgBoss.work(worker.queue, worker.perform))
+    }
```
lucianak, line 49: why do we not directly return 

`this.pgBoss.work(worker.queue, worker.perform)`?

packages/backend-modules/job-queue/lib/queue.ts
```diff
@@ -0,0 +1,95 @@
...
+    await this.pgBoss.start()
+  }
+
+  async stop() {
+    await this.pgBoss.stop({
+      timeout: 20000,
+    })
+
+    await new Promise((resolve) => {
+      this.pgBoss.once('stopped', () => resolve(null))
```
lucianak, line 32: if `pgBoss.stop` times out, will this Promise resolve properly?

lucianak (commented):
LGTM

packages/backend-modules/job-queue/lib/queue.ts
hdahlheim: if we would await each worker, the processing won't happen concurrently because each loop iteration would wait for the worker to start before starting the next one. By using `Promise.all` the workers get started concurrently and we can await all Promises at once.

PR #876, opened 2024-06-17
feat(www): allow redirecting to a set of white-listed hostnames after login

trm217:


apps/www/pages/anmelden.js
```diff
@@ -20,14 +23,23 @@ const SigninPage = () => {
...
+        if (allowedSignInRedirectHosts.includes(redirectUrl.host)) {
+          // redirect to the target URL using the browser's native redirect
+          window.location.assign(redirectTarget)
+          return
+        }
+
+        // ensure that the redirect target can only be a relative path
+        throw new Error('Invalid redirect URL')
       } catch (e) {
         router.replace('/')
```
jstcki, line 44: Again, I think I'd prefer not to use the router here at all.

apps/www/.env.example
```diff
@@ -14,6 +14,8 @@ NEXT_PUBLIC_BASE_URL=http://localhost:3010
 NEXT_PUBLIC_PUBLIKATOR_BASE_URL=http://localhost:3005
 NEXT_PUBLIC_ADMIN_BASE_URL=http://localhost:3006
 
+NEXT_PUBLIC_ALLOWED_SIGNIN_REDIRECT_HOSTS=localhost:3010,localhost:3005,localhost:3006,localhost:3000
+¨
```
jstcki, line 18: Should we use origin (incl. protocol) here instead of just the host?

jstcki (approved):

PR #877, opened 2024-06-19
fix(signin): insert login link as raw string

hdahlheim:
Inserts links as raw string to prevent double encoding of links


hdahlheim:
tested locally and on love

lucianak (approved):
👍

PR #879, opened 2024-06-19
refactor(www): remove access grant share chart

jstcki:
It's not used anywhere.

Let's check if the queries can be cleaned up also


lucianak (approved):
Looks good and works locally on my machine but let's also test on love to be sure 😁

hdahlheim (approved):

PR #886, opened 2024-07-02
feat(www): hook up cockpit to DatoCMS

trm217:


ovbm (approved):

apps/www/graphql/republik-api.schema.graphql
```diff
@@ -100,6 +100,59 @@ type AccessGrantInfo {
   message: String
 }
 
+type AccessGrantStats {
```
jstcki, line 103: These types were removed a while ago. Have you generated the schema from apo-prod? Maybe api-love is outdated?

PR #888, opened 2024-07-08
chore(base/graphql): set apollo server cache to bounded

hdahlheim:
The default cache config of apollo server can be used for DOS attacks


lucianak (approved):
👍

PR #898, opened 2024-08-26
Open Access Test

jstcki:
To make it work, also do these:

- (Completely) remove the `DOCUMENTS_RESTRICT_TO_ROLES` env var on the API
- Set `meta.isRestricted` field on the Front document to false (via Publikator)


jstcki:
Superseded by #902

PR #908, opened 2024-09-25
fix: redirect back to home after a redirect in the app

jstcki:
Sometimes, `history.back()` doesn't do anything (e.g. when the app is loaded fresh), so users can get stuck on the redirection page.


jstcki (commented):
Approved by @ovbm IRL 😉

PR #915, opened 2024-10-02
fix: swap usericon for login prompt

ovbm:
before
Screenshot 2024-10-02 at 17 41 22

after
Screenshot 2024-10-02 at 17 40 03

### DIFF
--- a/apps/www/components/Frame/User.js
+++ b/apps/www/components/Frame/User.js
@@ -28,7 +28,7 @@ export const SignInLink = ({ isOnMarketingPage }) => {
         <IconAccountBox {...colorScheme.set('fill', 'text')} />
       </span>
       <span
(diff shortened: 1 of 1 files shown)


apps/www/components/Frame/User.js
```diff
@@ -28,7 +28,7 @@ export const SignInLink = ({ isOnMarketingPage }) => {
         <IconAccountBox {...colorScheme.set('fill', 'text')} />
       </span>
       <span
-        {...(isOnMarketingPage ? styles.labelMarketing : styles.labelDefault)}
+        {...styles.label}
```
ovbm, line 31: changed it to just one label css because labelMarketing and labelDefault were the exact same css XD

jstcki (approved):

PR #917, opened 2024-10-03
fix(SyntheticReadAloud): associate derivatives with commits so that not every publish or commit needs a newly generated audio if not necessary

lucianak:
add feature to skip audio generation while publishing, and use audio from latest published commit instead. Adds new table to associate derivatives with commits.

### DIFF
--- a/packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
+++ b/packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
@@ -242,7 +242,6 @@ export const derive = async (
       WHERE status = 'Pending'
       AND c."repoId" = :repoId;`, {repoId: document.repoId})
       
-      console.log('pending count: %s', pendingCount)
     if (pendingCount > 0) {
       handlerDebug('more than one pending derivatives for this repo. skipping synthesizing.', {
         userId: user.id,
--- a/packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
+++ b/packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
@@ -161,6 +161,11 @@ export const onPublish = async (
       { repoId: repoId },
     )
 
+    if (!latestPublishedDerivative) {
+      console.error('No previously published successful derivative available, not associating derivative with commit.')
+      return
+    }
+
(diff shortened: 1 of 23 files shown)


packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
```diff
@@ -131,6 +140,32 @@ export const onPublish = async (document: any, pgdb: any, user?: any) => {
...
+        WHERE repos."currentPhase" in ('scheduled', 'published')
+        AND milestones.scope = 'publication'
+        AND d.status = 'Ready'
+        AND d.type = 'SyntheticReadAloud'
+        AND repos.id = :repoId
+        ORDER BY milestones."publishedAt" DESC;`,
+      { repoId: repoId },
+    )
+
+    await associateReadAloudDerivativeWithCommit(latestPublishedDerivative, commit, pgdb)
```
annatraussnig, line 164: do we handle the case where there is not previous derivative available?

hdahlheim (approved):
👍 Nothing to add

packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
```diff
@@ -164,22 +236,20 @@ export const derive = async (
...
-    //  (probably remove altogether when we delete the 'audio generieren' button)
-    if (pendingCount > 1) {
-      handlerDebug('too many pending derivatives. skipping synthesizing.', {
+    const pendingCount = await pgdb.queryOneField(`SELECT COUNT(d.id)
+      FROM publikator.derivatives d
+      JOIN publikator.commits c ON c.id = d."commitId"
+      WHERE status = 'Pending'
+      AND c."repoId" = :repoId;`, {repoId: document.repoId})
+      
+      console.log('pending count: %s', pendingCount)
```
annatraussnig, line 245: is this meant to stay?

annatraussnig (approved):

packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
lucianak: oops

PR #922, opened 2024-10-08
Include new subscriptions in cockpit query

lucianak:
Correct data for charts on [cockpit](https://www.republik.ch/cockpit) so that new members with subscriptions are correctly shown. 
Relevant data:
- active subscribers
- new subscriptions, ended subscriptions
- subscribers active on the site
- discussions: doesn't depend on membership type


hdahlheim (approved):
Thank you for the work! Let's hope we can get rid of this mess at some point 😁

PR #924, opened 2024-10-10
Voices

annatraussnig:
Add a mark button in publikator that allows tagging a section of text with a different voice than the one from the metadata. 

https://github.com/user-attachments/assets/5d6380b4-d3a4-4309-929c-729af1f63354

### DIFF
--- a/apps/publikator/components/editor/index.js
+++ b/apps/publikator/components/editor/index.js
@@ -15,6 +15,7 @@ import createBlockTextModule from './modules/blocktext'
 import createLinkModule from './modules/link'
 import createMarkModule from './modules/mark'
 import createMemoModule from './modules/memo'
+import createVoiceModule from './modules/voice'
 import createListModule from './modules/list'
 import createListItemModule from './modules/list/item'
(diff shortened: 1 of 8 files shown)


lucianak (commented):
Looks great! I can't say much about the frontend code though :) 
Some questions: how does the markdown look with the voice tags? Are we checking that edge cases like people marking parts of a paragraph or the title in a different voice don't break the publikator? Can you combine voice tagged paragraphs with memos?

annatraussnig:
@lucianak all excellent questions. In order:

1. we use `<span>` tags with a `data-voice` attribute to allocate segments of text to a different voice.
2. we are currently not checking edge cases. the idea was to first make it work. 
3. combining voice tags with memos wouldn't work, the voice tags would have to be added towards the end of the production process, once the memos are resolved.

In short, it's not the most stable feature yet, but used responsibility, it works. As we are still on the fence regarding whether we want multivoice texts or not, I would postpone looking into those (very much not-imaginary) questions.

Converting PR into draft.

PR #929, opened 2024-10-14
chore(calculateKpis): adapt to subscriptions

lucianak:
The subscriptions should be included with the memberships and abonnements, and marked as reduced if a coupon was applied. Hence there are no new fields for PROJECT_R,  because they are included with ReduzierteMitgliedschaften, but new fields for REPUBLIK, for ReduzierteAbonnements and StornierteReduzierteAbonnements. The payment method is still Stripe, but might change when we start using Stripe with Paypal/Twint, depending on how the booking is organised between stripe and external payment providers.

### DIFF
--- a/packages/backend-modules/republik/script/finance/calculateKpis.js
+++ b/packages/backend-modules/republik/script/finance/calculateKpis.js
@@ -39,6 +39,10 @@ const currency = new Intl.NumberFormat('de-CH', {
   useGrouping: false,
 })
 
+const isCancelledOrRefunded = (item) => {
+  return ['CANCELLED', 'REFUNDED', 'FAILED'].includes(item.status.toUpperCase())
+}
+
 const evaluateCompanyMonth = async (
   company,
   begin,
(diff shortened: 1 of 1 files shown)


packages/backend-modules/republik/script/finance/calculateKpis.js
```diff
@@ -135,7 +166,7 @@ const evaluateCompanyMonth = async (
         .filter((i) => i.companyName === company)
         .filter((i) => i.method === method)
         .filter((i) => i.type === 'MembershipType')
-        .filter((i) => ['CANCELLED', 'REFUNDED'].includes(i.status))
+        .filter((i) => ['CANCELLED', 'REFUNDED', 'failed', 'refunded'].includes(i.status))
```
hdahlheim, line 169: It would make sense define this filter function as a named function and use that instead of redefining six times.
I am also for normalising the case of `i.status` with `i.status.toUpperCase()` so that we don't have duplicated values.

PR #931, opened 2024-10-15
feat: add more granular slow speeds

ovbm:
Users requested more granluar playback rates, which is easy to do. 

Tested locally.


annatraussnig (approved):

lucianak (approved):
Looks good 👍 thanks for doing it

PR #933, opened 2024-10-15
fix(editorCore): forbid consecutive newline inserts

annatraussnig:
Consecutive newlines in the title block are the ex-aequo number one cause of editor bugs in publikator (together with scrambled memos), as markdown interprets two consecutive `\n\n` as a new paragraph. However, multiple titles, or leads, are not tolerated.

This PR adds two new behaviours:

1. Ignore every newline insert that follows or precedes another newline character (`\n`)
2. Collapse multiple consecutive newlines into a single newline on copy-paste. An exception is made for paragraph types, where multiple newlines are converted into a double newline, which leads to the creation of a new paragraph (desired behaviour for paragraph types)

### DIFF
--- a/apps/publikator/components/editor/modules/document/createPasteHtml.js
+++ b/apps/publikator/components/editor/modules/document/createPasteHtml.js
@@ -18,12 +18,21 @@ const hasParent = (type, document, key) => {
 const normalise = (html) =>
   html.replace(/<b[^>]*font-weight\s*:\s*normal[^>]*>/g, '')
 
+const PARAGRAPH_TYPES = ['PARAGRAPH', 'INFOP', 'BLOCKQUOTEPARAGRAPH']
+
 export default (centerModule, figureModule) => (event, change, editor) => {
   const transfer = getEventTransfer(event)
-  if (transfer.type !== 'html') return
-
   const cursor = editor.value.selection.anchorKey
   const blockType = editor.value.document.getClosestBlock(cursor).type
+
+  if (transfer.type !== 'html') {
+    if (PARAGRAPH_TYPES.includes(blockType)) {
+      change.insertText(transfer.text.replace(/\n{2,}/g, '\n\n'))
+      return true
+    }
+    change.insertText(transfer.text.replace(/\n+/g, '\n'))
+    return true
+  }
 
   const isByline = blockType === 'CENTERBYLINE' || blockType === 'BYLINE'
   if (isByline) return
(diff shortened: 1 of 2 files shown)


apps/publikator/components/editor/utils/keyHandlers.js
```diff
@@ -158,9 +159,12 @@ export const createSoftBreakKeyHandler =
     if (event.key !== 'Enter') return
     if (event.shiftKey === false) return
 
-    const { startBlock } = value
+    const { startBlock, startText, endOffset } = value
+
+    const isConsecutiveSoftBreak = startText?.text[endOffset - 1] === '\n' || startText?.text[endOffset] === '\n'
```
jstcki, line 164: I wonder if instead of using the offset directly, we could just check `startText.endsWith('\n')` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith ? Or is the `endOffset` not always the end of the string? Anyway, not that important.

jstcki (approved):
Looks good to me but I don't really understand the semantics of `startText` etc.

I wonder what happens when you paste text with multiple newlines though?

apps/publikator/components/editor/utils/keyHandlers.js
annatraussnig: ah, excellent question i didn't think of that... i'll find a solution for that, too

annatraussnig: thanks for the `startText` suggestion, it's a good question too and I don't know the answer, so I'll try that as well.

annatraussnig: I checked the docs: `startText` is the whole text node where the `startCursor` is placed, while `endText` is the text node of the `endCursor`. `startOffset` and `endOffset` play a similar role for text offsets. If if the selection is collapsed (no text is selected and the cursor is just blinking) start and end cursors are the same. I changed the code so that things also work in the case the selection is active.

jstcki (approved):

apps/publikator/components/editor/modules/document/createPasteHtml.js
```diff
@@ -18,13 +18,22 @@ const hasParent = (type, document, key) => {
...
 export default (centerModule, figureModule) => (event, change, editor) => {
   const transfer = getEventTransfer(event)
-  if (transfer.type !== 'html') return
-
   const cursor = editor.value.selection.anchorKey
   const blockType = editor.value.document.getClosestBlock(cursor).type
 
+  if (transfer.type !== 'html') {
+    if (PARAGRAPH_TYPES.includes(blockType)) {
+      change.insertText(transfer.text.replace(/\n{2,}/g, '\n\n'))
```
jstcki, line 30: @annatraussnig re the sentry error: maybe this needs to be `transfer.text?.replace` (also below)?

annatraussnig: yep i merged a fix for this.

PR #934, opened 2024-10-15
Update deleteUser and mergeUser mutations for new subscriptions

lucianak:
- update deleteUser to work with new subscriptions
- update mergeUser to work with new subscriptions
- disallow deleting your own user
- disallow deleting a user with an active membership or subscription

### DIFF
--- a/packages/backend-modules/republik-crowdfundings/lib/payments/stripe/deleteCustomer.js
+++ b/packages/backend-modules/republik-crowdfundings/lib/payments/stripe/deleteCustomer.js
@@ -11,11 +11,14 @@ module.exports = async ({ userId, pgdb }) => {
     return
   }
 
-  for (let customer of customers) {
+  for (const customer of customers) {
     const account = accounts.find((a) => a.company.id === customer.companyId)
     if (account) {
       await account.stripe.customers.del(customer.id)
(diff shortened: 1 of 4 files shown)


packages/backend-modules/republik-crowdfundings/lib/payments/stripe/deleteCustomer.js
```diff
@@ -11,13 +11,16 @@ module.exports = async ({ userId, pgdb }) => {
...
-  for (let customer of customers) {
+  for (const customer of customers) {
     const account = accounts.find((a) => a.company.id === customer.companyId)
     if (account) {
       await account.stripe.customers.del(customer.id)
       await pgdb.public.stripeCustomers.deleteOne({
         id: customer.id,
       })
+      await pgdb.payments.stripeCustomers.deleteOne({
+        id: customer.id,
```
hdahlheim, line 22: this should be customerId: customer.id

hdahlheim: but in the end we should just get rid of the new stripe table. When I drafted the payments module I would it would be simpler to kept in the same db schema but I forgot that the old order flow also needs them.

PR #935, opened 2024-10-21
feat(MultiVoice): add voice module

annatraussnig:
Add an interview-answer button that wraps around paragraphs.
> Screenshot 2024-10-21 at 12 06 16

Whatever is marked as interview-answer gets spoken with `preselected voice 2`:
> Screenshot 2024-10-21 at 12 07 38

This iteration happens at block level in the mdast tree, and is hence more robust than the marking previously proposed.

Todo:

- [x]  remove margin between question and answer:
> Screenshot 2024-10-21 at 12 09 15


lucianak (commented):
Looks good, same here that I think it would be good to have a warning/error if the secondary voice is not selected and the blocks are used so we don't accidentally regenerate the whole article if it's forgotten :)

annatraussnig:
The error is throw (and propagated) by the tts server. since it's just for interviews that feels enough to me – what do you think? do you think a warning on publish is also necessary?

PR #937, opened 2024-10-22
fix: remove trial signup in audio player

ovbm:
Maintenance Task

Next up (in seperate pr): consider paynote height for audio-player postition, so that it doesn't overlap.

Screenshot 2024-10-22 at 11 27 29


annatraussnig (approved):
looking good :) 
tested locally

PR #856, opened 2024-05-30
WIP: Print CSS improvements

ovbm:

PR #938, opened 2024-10-22
feat(print): adds media print queries to paynotes, headers, collapsables

ovbm:
Should make our default print view more consistent, so we could in theory deprecated react-pdf in the future. Cherry picked a commit I did 4 months ago:  https://github.com/republik/plattform/pull/856/commits/49a97577b5bffa1ada2f686f530df688e66144c5

We should merge as there's no downside in adding the styles (all the changes only applie to media:print) and having it laying around is not a good idea. Deleted the old PR. 

ToDO
- [ ] Test on Love


annatraussnig (approved):
Looks good (not tested on love yet)

PR #939, opened 2024-10-22
fix: consistent audio player offset & env(safe-area-) fixes

ovbm:
This PR fixes `env(safe-area-inset-...)` not working properly. It wasn't working in the webview of our react native app, because `viewport-fit=cover'` wasn't set on the meta tag.

Now that it's set it is working as expected and the Callout Element has been simplified by removing the explicit definition of mobile paddings. Now it defines consistent paddings that work both in Native and Web views

example web:
Screenshot 2024-10-22 at 15 19 53
same code in native (notice that the marginBottom is larger, because the bottom margin is added to the safe area start `calc(15px + env(safe-area-inset-bottom))`. This ensures consistent hit boxes and enough space that the native UI elements and our buttons don't conflict:
Screenshot 2024-10-22 at 15 19 45

before the change, the last item of the list was very close to the bottom UI
Screenshot 2024-10-22 at 15 16 19

using env(safe-are-inset..) ensures that the additonal paddings are only ever applied in the native app, since browsers on the web or mobile don't have safe-area-insets.

 new consistent audio-player paddings:
Screenshot 2024-10-22 at 15 04 18

### DIFF
--- a/apps/www/components/Article/ActionBarOverlay.js
+++ b/apps/www/components/Article/ActionBarOverlay.js
@@ -14,9 +14,7 @@ const ActionBarOverlay = ({ children }) => {
   const isDesktop = useMediaQuery(mediaQueries.mUp)
   const { audioPlayerVisible } = useAudioContext()
 
-  const audioPlayerOffset = audioPlayerVisible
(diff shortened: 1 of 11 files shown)


annatraussnig (approved):
⭐

PR #940, opened 2024-10-22
Datawrapper

annatraussnig:
What work:

- [ ] sidebar `datawrapper` button in publikator
- [ ] modal to input dw id & alt text
- [ ] md/mdast/schema stuff in place
- [ ] chart shows up in the preview iframe
- [ ] code looks ok

What doesn't work:

- [x] chart no display in publikator
- [x] chart flickers a brief instant into existence on www, then disappears

... Upon inspection, code looks as it should (same as embed code) though, but a priori this needs some tuning still...

Motivational pictures

Screenshot 2024-10-22 at 17 56 18

Screenshot 2024-10-22 at 17 48 33

### DIFF
--- a/apps/publikator/components/editor/index.js
+++ b/apps/publikator/components/editor/index.js
@@ -52,11 +52,13 @@ import createVariableModule from './modules/variable'
 import createVariableConditionModule from './modules/variable/condition'
 import createSeriesNavModule from './modules/seriesnav'
 import createAuthorModule from './modules/author'
+import createEmbedDatawrapperModule from './modules/embeddatawrapper'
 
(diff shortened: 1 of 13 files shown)


annatraussnig:
Ah very good! The edge-to-edge thing requires us to slice the mdast tree etc., if we decide we don't want it we can simplify this part of the publikator module. If we want it, we need to specify in the schema that DW be children of the root element. At the moment it's only allowed where paragraphs are allows (aka: not as children of `root`).
I could see how for a more explorative dataviz, edge-to-edge could be nice, but maybe this pandora doesn't want to open that one box yet.

PR #946, opened 2024-10-30
feat: wip allowIP

ovbm:


apps/www/middleware.ts
```diff
@@ -233,10 +234,46 @@ async function middlewareFunc(req: NextRequest): Promise<NextResponse> {
...
+    token: string,
+  ): Promise<NextResponse> {
+    try {
+      // Parse and verify JWT to decide about redirection
+      const jwtBody = await verifyJWT(token)
+      const clientIP = jwtBody?.ip
+      const isAllowedIP =
+        clientIP &&
+        process.env.IP_ALLOWLIST &&
+        process.env.IP_ALLOWLIST.includes(clientIP)
```
annatraussnig, line 252: Question: would there be only one IP address allowed per customer? do we need to support wildcards and/or ranges here?

ovbm: Good question. Currently we have one request by Swissconsortium that uses one specific IP per customer. I would say we cross the ip-ranges and wildcards if there is a need for it.

apps/www/pages/_app.tsx
```diff
@@ -64,8 +67,29 @@ const WebApp = ({
   } = pageProps
 
   const router = useRouter()
+  const [allowlistAccess, setAllowlistAccess] = useState(false)
```
ovbm, line 70: This is probably not the best place to have this useEffect. Should be in its own hook or incorporated somewhere else. I just put it here because this is where we decide to show the paynotes - or not.

PR #950, opened 2024-11-05
Kill backend styleguide

annatraussnig:
This pull request move the templates used to generate email content into its own package:
`mdast-mail-templates`

Why in the mdast folder and not together with the backend module?

1. Rendering isn't really a backend responsibility
2. The templates take an mdast tree as an input
3. The template are always used in conjuction with another libbrary, mdast-react-render. They belong together.

What to test:

[ ] email with a comment
[ ] article as email
[ ] newsletter as email
[ ] comment rendering in www (since we deleted some files)
[ ]  article rendering in www
[ ] newsletter rendering in www

If all those things are working, it is safe to merge this PR.


packages/mdast/mail-templates/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js",
+      "import": "./dist/index.mjs",
+      "types": "./dist/index.d.ts"
+    },
+    "./email": {
```
jstcki, line 15: I don't see the source file for this, we can remove it, no?

annatraussnig: excellent catch as always

PR #959, opened 2024-11-12
feat: ip blocklist middleware

ovbm:
This would allow us to define IP addresses to block in a comma seperated IP_BLOCKLIST env variable

### DIFF
--- a/apps/www/middleware.ts
+++ b/apps/www/middleware.ts
@@ -133,13 +133,20 @@ async function middlewareFunc(req: NextRequest): Promise<NextResponse> {
   }
 
   const resUrl = req.nextUrl.clone()
+
   // Block if request is coming from IP_BLOCKLIST
   const clientIp = req.headers['x-forwarded-for']
     ? req.headers['x-forwarded-for'].split(',')[0]
     : ''
 
-  if (clientIp === '129.80.232.132') {
-    console.log(`request with IP: ${clientIp} denied.`)
+  const isBlacklistedIP =
+    clientIp &&
+    process.env.IP_BLOCKLIST &&
+    (process.env.IP_BLOCKLIST || '')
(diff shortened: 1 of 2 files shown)


apps/www/middleware.ts
```diff
@@ -133,13 +133,20 @@ async function middlewareFunc(req: NextRequest): Promise<NextResponse> {
...
     ? req.headers['x-forwarded-for'].split(',')[0]
     : ''
 
-  if (clientIp === '129.80.232.132') {
-    console.log(`request with IP: ${clientIp} denied.`)
+  const isBlacklistedIP =
+    clientIp &&
+    process.env.IP_BLOCKLIST &&
+    (process.env.IP_BLOCKLIST || '')
+      .split(',')
```
annatraussnig, line 146: i am not sure about the splitting. i feel like this also adds potential for erros if someone adds a whitespace after the comma or something. 
i think i would do:
`process.env.IP_BLOCKLIST.includes(clientIp)` without the splitting. Since IPs don't include commas this should work too, no?

apps/www/middleware.ts
```diff
@@ -133,13 +133,20 @@ async function middlewareFunc(req: NextRequest): Promise<NextResponse> {
...
+      .some((ip) => clientIp.includes(ip))
+
+  if (isBlacklistedIP) {
```
annatraussnig, line 149: do we need a more professional server response or we dgaf?

ovbm: DGAF!!! :D

apps/www/middleware.ts
ovbm: i see what you mean. makes sense

annatraussnig (approved):

PR #960, opened 2024-11-13
fix: remove twitter handle from footer, add webshare api

ovbm:
- Removes Twitter from our footer 
- uses web Share API instead of our own share overlay if a mobile device (ios or android) is used

tested on love


apps/www/components/ActionBar/ShareButtons.js
```diff
@@ -46,33 +46,6 @@ const ShareButtons = ({
     }
   }, [copyLinkSuffix])
```
ovbm, line 48: The only place this was ever used was the end of a series overview. I removed that so the ShareButtons component now only ever renders the sharebuttons without a native app fallback.

apps/www/components/ActionBar/index.js
```diff
@@ -382,20 +404,9 @@ const ActionBar = ({
...
-            payload: {
-              title: document.title,
-              url: shareUrl,
-              subject: emailSubject,
-              dialogTitle: t('article/share/title'),
-            },
-          })
+        handleShareClick()
+        if (isNativeApp) {
           e.target.blur()
```
annatraussnig, line 409: it kinda bothers me that the if clause in the `handleShareClick` function is the same as the one to blur the event target (`if (inNativeApp)`).
I would pass the event/event.target to `handleShareClick`

annatraussnig (approved):
Code looks good. Didn't test on mobile.

apps/www/components/ActionBar/index.js
ovbm: yeah i passed it in before, don't know why i removed it on a later refactor...

ovbm: done

PR #961, opened 2024-11-19
Datawrapper email

annatraussnig:
- add chart capabilities to article-as-newsletters (e.g. wdwww) and editorial newsletters
- charts are rendered as 450px wide screenshots

Editorial newsletter:
Screenshot 2024-11-19 at 17 11 46

Article-as-newsletter:
Screenshot 2024-11-19 at 17 21 50

### Article as Newsletter

Any chart in the article will also be sent as a picture via email automatically.

### Editioral Newsletter

Default to plain chart (no header or footer, since those feel more article-y).


annatraussnig:
Yeah mobile is a rather small. I think I'll try to change the default width from 600px to 450px.
IMG_C3F588741675-1

apps/publikator/components/editor/modules/embeddatawrapper/index.js
```diff
@@ -62,7 +62,10 @@ const EmbedDatawrapper = ({ rule, TYPE }) => {
   const newItem = () =>
     Block.create({
       type: TYPE,
-      isVoid: true
+      isVoid: true,
+      data: {
+        plain: !resizable,
```
jstcki, line 67: What does this do? Isn't `plain` its own option, unrelated to resizable?

packages/mdast/mail-templates/src/newsletter/schema.jsx
```diff
@@ -371,31 +260,22 @@ const createNewsletterSchema = ({
...
-                  insertButtonText: 'Trennlinie',
-                  insertTypes: ['PARAGRAPH'],
-                },
-                isVoid: true,
               },
             ],
           },
           {
             matchMdast: () => false,
-            editorModule: 'specialchars',
```
jstcki, line 398: What are all these deletions? Are they not needed anymore?

packages/mdast/mail-templates/src/shared/components/Datawrapper.jsx
```diff
@@ -0,0 +1,44 @@
...
+    src={src}
+    alt={alt}
+  />
+)
+
+export const Datawrapper = ({ datawrapperId, alt, plain }) => (
+  <Figure>
+    <Chart
+      src={`https://datawrapper.dwcdn.net/${datawrapperId}/mail${plain ? 'plain' : ''}.png`}
+      width={450}
```
jstcki, line 30: In the [figureRule](https://github.com/republik/plattform/pull/961/files#diff-ba4293d866802fc6f9ee5581a78e8aa12b602cd8eead03a61f08ca60b1236d69R32) it's defined as 600px?

packages/mdast/mail-templates/src/shared/components/Datawrapper.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { EdgeToEdgeFigure, Figure } from './Figure'
+
+// We generate the following images in Datawrapper upon publishing:
+//  1. https://datawrapper.dwcdn.net/DW123ID/mail.png -> 450px wide
+//  2. https://datawrapper.dwcdn.net/DW123ID/mailplain.png -> 450px wide, no header or footer
+//  3. https://datawrapper.dwcdn.net/DW123ID/edgetoedge.png -> 1280px wide
+//  4. https://datawrapper.dwcdn.net/DW123ID/edgetoedgeplain.png -> 1280px wide, no header or footer
+
+const Chart = ({ src, alt, width }) => (
```
jstcki, line 10: Where does the alt prop come from? It's part of the chart metadata but we don't load that anywhere.

packages/styleguide/src/templates/shared/rules/embedDatawrapperRule.js
```diff
@@ -0,0 +1,21 @@
+import { matchZone } from '@republik/mdast-react-render'
+import Datawrapper from '../../../components/Datawrapper'
+
+export const embedDataWrapperRule = (resizable = true) => ({
```
jstcki, line 4: Can we make that an object, so it becomes `embedDataWrapperRule({ resizable: false })` to make it more obvious what this argument does?

jstcki (commented):
Super-nice 🌟

Are articles-as-emails also covered here?

I was wondering if the `resizable` prop is necessary to have as a rule option. Can't we just ignore the size in the email templates? I believe if we send articles as emails this needs to be the case anyway?

packages/mdast/mail-templates/src/newsletter/schema.jsx
annatraussnig: ah, the mail templates actually don't need the editorModule/editorOptions since we only use them to render emails, not to work in Publikator. (We use the templates from styleguide, and those since have the editorModule/editorOptions).
So I deleted them. Should do this for all the mdast/mail-templates.

packages/mdast/mail-templates/src/shared/components/Datawrapper.jsx
annatraussnig: yes, exactly. i tried sending a test email with width 600px and the labels were so tiny they were hurting my eyes. so i wanted to a width of 450px. Not sure… second opinion appreciated on this.

packages/styleguide/src/templates/shared/rules/embedDatawrapperRule.js
annatraussnig: we can. regarding your question with respect to size/resizing. it is allowed for figures in the article-as-email template, which is why i decided to keep it consistent and allow it there. Editorial newsletter don't allow resizable charts.

packages/mdast/mail-templates/src/shared/components/Datawrapper.jsx
annatraussnig: ah, we removed it from the config modal. what's your opinion? re-add it there or delete it here? i think it's kinda nice to have but i just saw that DW also has a Beschriebung field, so maybe let's just delete it everywhere?

packages/styleguide/src/templates/shared/rules/embedDatawrapperRule.js
jstcki: Yes, when the charts are rendered in the browser, DW takes care of including the alt text. That's why I removed it from the modal. To use the same text on the images would require us to load the embed.json of the chart and set it on the image … not sure where to do that in this module maze.

packages/mdast/mail-templates/src/shared/components/Datawrapper.jsx
jstcki: OK, yes, makes sense. I was just checking because I saw two different numbers. So the 450px image will not be scaled up to 600px in the email (I have no idea what the image size in the figureRule does)?

packages/styleguide/src/templates/shared/rules/embedDatawrapperRule.js
jstcki: OK, we can also keep it. And in the article emails the size will be ignored?

packages/mdast/mail-templates/src/newsletter/schema.jsx
jstcki: Nice!

packages/mdast/mail-templates/src/shared/components/Datawrapper.jsx
annatraussnig: removed alt prop

packages/styleguide/src/templates/shared/rules/embedDatawrapperRule.js
annatraussnig: editorial newsletter, no resize:
- single sized chart (450 or 600px, depending what we decide)

article emails, two sizes for charts:
- regular chart (450/600px wide)
- edge-to-edge (like article-email figure, 1280px wide)

packages/mdast/mail-templates/src/shared/components/Datawrapper.jsx
annatraussnig: no – then the max width for the chart-image is 450px. as i said, i am a bit on the fence about which size a regular email chart should be...

apps/publikator/components/editor/modules/embeddatawrapper/index.js
annatraussnig: yeah, i just thought maybe the editorial newsletter, since the header/footer of the chart don't resize, they look a bit awkward, so not displaying them by default makes sense…? 🤷‍♂️ 
Unless you think it's a terrible idea in the first place, I'll add a comment in the code to clarify.

annatraussnig: refactored to make it clearer.

annatraussnig:
I think everything is clarified. I also sent out a test email with 450px as width for the regular charts, which I rather like.

jstcki (approved):
Let's gooo

PR #966, opened 2024-11-25
feat(payments/gifts): base of the new gifts for the new magazine subscriptions

hdahlheim:
Start of a new gift voucher system that works with subscriptions and pledges.


hdahlheim:
This PR refactors a lot of places to make gifts for the new subscriptions possible we the gifts are still gated behind a feature flag because we haven't decided the final flow. To prevent this branch from further diverging with main we decided to merge and finish the rest in a separate PR.

PR #968, opened 2024-11-27
Redirects /angebote to shop.republik.ch

ovbm:
This redirects blank requests to /angebote (without any query params) to shop.republik.ch. If query params `package`, `group` or `goto` are set, the redirect doesn't apply.


annatraussnig (approved):
Let's Get This Merged!
(Soon)

PR #971, opened 2024-12-02
fix(InterviewAnswer): replace wrapping div with css class

annatraussnig:
Replace wrapping `div` around interview-answer `p` with a `no-top-margin` class, in the hope it fixes our pdf problem.


jstcki (approved):
Nice!

PR #977, opened 2024-12-11
Fix Image Copy on Commit

annatraussnig:
**The problem:** when processing a document with loads of images to copy, too many requests were made at once. This caused the asset server to crash and the commit to fail.

**The solution:** we now copy images sychronously. It takes a lot longer, and the backend might randomly hang up, but it works most of the time.

### DIFF
--- a/packages/backend-modules/documents/lib/mdast/process.js
+++ b/packages/backend-modules/documents/lib/mdast/process.js
@@ -43,7 +43,7 @@ const processRepoImageUrlsInContent = async (content, fn) => {
     }
   })
 
-  return Promise.all(fns.map((fn) => fn()))
+  return Promise.each(fns, fn => fn())
 }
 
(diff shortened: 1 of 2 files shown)


packages/backend-modules/documents/lib/mdast/process.js
```diff
@@ -43,7 +43,7 @@ const processRepoImageUrlsInContent = async (content, fn) => {
     }
   })
 
-  return Promise.all(fns.map((fn) => fn()))
+  return Promise.each(fns, fn => fn())
```
annatraussnig, line 46: I also tried to limit the concurrency with:
`Promise.map(fns, fn => fn(), { concurrency: 5 }`
but the asset server was still crashing.

annatraussnig: @lucianak

PR #980, opened 2024-12-17
Generic(-ish) Questionnaire

annatraussnig:
### Idea

- generic overview and person view
- all configuration is extracted into… wait for it… a config file (this could also live in the DC props, but feels messier)
- config specifies the colors, illu, and questions grouping
- each new questionnaire gets its own config but that's it
- dynamic component specifies a config key (e.g. "klima-2023", or "democracy") which we use to retrieve the correct config
- form and overview are publikator pages with embedded QUESTIONNAIRE components (like for klimafragebogen)
- person page is a next page with 2 params, config key and user id (we could do something cute like: `{userslug}/aeussert-sich-ueber/${key}`, aka `annatraussnig/aeussert-sich-ueber/demokratie`)

### Discuss

- [ ] paragraphs between questions UI


apps/www/components/Questionnaire/Submissions/PersonViewHeaderShare.js
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <>
+      <IconButton
+        label={!noLabel && t('article/actionbar/share')}
+        labelShort=''
+        Icon={IconShare}
+        href={url}
+        onClick={(e) => {
+          e.preventDefault()
+          trackEvent(['ActionBar', 'shareQuestionnaireLink', url])
```
jstcki, line 28: Need to check myself, but does this still work with Plausible? Probably better to use the useTrackEvent hook

apps/www/components/Questionnaire/Submissions/graphql.js
jstcki, line 1: Can we, should we move these over to .gql files in graphql/republik-api/queries? That way the TS types would be generated and it would make it a bit more safe to use the results?

apps/www/components/Questionnaire/TextQuestion.js
```diff
@@ -47,13 +47,15 @@ class TextQuestion extends Component {
...
-      question: { text, explanation, maxLength },
+      question: { text, explanation, maxLength, order, metadata },
       t,
     } = this.props
     const { value } = this.state
+    const beforeP = metadata && metadata.beforeP
     return (
       <div>
-        {text && <H2 {...questionStyles.label}>{text}</H2>}
+        {beforeP && <Editorial.Subhead>{beforeP}</Editorial.Subhead>}
```
jstcki, line 57: Will we understand what `beforeP` does next year still? 😅

apps/www/pages/fragebogen/[slug]/[id].js
Comment (jstcki) [inline apps/www/pages/fragebogen/[slug]/[id].js:1]: Should we name the params a bit more explicitly? `slug` is fine but maybe `userId`?

apps/www/tsconfig.json
```diff
@@ -15,8 +15,9 @@
       ],
       "#graphql/*": [
         "./graphql/*"
-      ],
-    }
+      ]
+    },
+    "jsx": "preserve"
```
jstcki, line 20: Why this change?

apps/www/components/Questionnaire/enhancers.js
jstcki, line 1: Same as above, should these queries be moved to .gql files?

jstcki (commented):
First glance looks good. I find the naming of things a bit confusing (slug/dbSlug/configId), not sure I understand the difference from just looking at the code. I realize this is a leftover from the old questionnaire code but wonder if we could clean this up.

Another thing I wonder: why do we even need Publikator pages (especially for the overview …) since we have a lot of content already in the config for the person answer view?

Is the idea to migrate the previous questionnaires to this structure as well?

apps/www/components/Questionnaire/TextQuestion.js
annatraussnig: oh ye of little faith

annatraussnig:
Thanks for the thoughts… 
Regarding setting up the overview as a page: I feel like everything that **can** be in the CMS **should** be in the CMS. The only reason the person's page is hardcoded is because it doesn't quite work otherwise to generate a new page for each person.
Migrating other quizzes: that's an excellent point, and one that my subconscious chose to block out til now, so thanks for bringing out to the light. I think we should be able to migrate all the questionnaires, save for the one with the politicians answer questions. That one is set up a bit differently. That said, it'd prefer a two-steps approach, i.e.: first merge the present PR, then open a new PR where we migrate the old questionnaires.
I'll refactor the names as much as I'm able to and migrate the graphql queries.

apps/www/components/Questionnaire/TextQuestion.js
annatraussnig: renaming to `questionHeader`

apps/www/tsconfig.json
annatraussnig: mistake – reverted.

apps/www/pages/fragebogen/[slug]/[id].js
Comment (annatraussnig) [reply apps/www/pages/fragebogen/[slug]/[id].js:1]: renamed to `authorId`

annatraussnig:
update on the graphql stuff – just spent some time on it. there are a lot of queries used in a lot of different ways. imo migrating and testing would entail a long time and only makes sense in the context of a big cleanup/refactoring of the questionnaire folder. so i'd flag as out of scope for now, but open to counter arguments :)

apps/www/components/Questionnaire/QuestionHeader.js
```diff
@@ -0,0 +1,23 @@
...
+  const questionHeader = metadata && metadata.questionHeader
+  if (!questionHeader) {
+    return null
+  }
+  return (
+    <div {...questionStyles.subheadOuter}>
+      <Editorial.Subhead {...questionStyles.subhead}>
+        <span
+          {...questionStyles.subheadInner}
+          style={{ backgroundColor: '#D9F0EF' }}
```
jstcki, line 14: Is this intended to stay?

image

jstcki (approved):
Code and Preview look good (one mini-WIP-leftover?)! Agree on doing the more thorough cleanup things in another PR but I hope we can get to it soon, lest we forget about it again 😉

apps/www/components/Questionnaire/QuestionHeader.js
annatraussnig: this is was a suggestion from vivienne and is in discussions with silvia...

jstcki: Yes, I'm not against it but the padding at the very beginning looks a bit odd, and we might want to reuse an existing color?

PR #981, opened 2024-12-18
fix: remove pledge statements

ovbm:


jstcki (approved):
Haven't tested but code LGTM

PR #984, opened 2024-12-20
chore: Remove unused social media fields

ovbm:
Removes twitterHandle, facebookId and publicUrl from front- and backend. 
 
ToDo:
- [ ] Test migration and deployment on staging


packages/backend-modules/auth/migrations/sqls/20241220130020-removeTwitterFacebook-up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE users
+DROP COLUMN "twitterHandle";
+DROP COLUMN "facebookId";
+DROP COLUMN "publicUrl";
```
hdahlheim, line 4: same here, `DROP` statements need to be separated by commas

packages/backend-modules/auth/migrations/sqls/20241220130020-removeTwitterFacebook-down.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE users
+ADD COLUMN "publicUrl" text;
+ADD COLUMN "facebookId" text;
+ADD COLUMN "publicUrl" text;
```
hdahlheim, line 4: The SQL syntax here is wrong. `;` ends a statement so the `ADD COLUMN` after `publicUrl` will fail.

The lines should be delimited by `,`

packages/backend-modules/auth/migrations/sqls/20241220130020-removeTwitterFacebook-up.sql
ovbm: done

packages/backend-modules/auth/migrations/sqls/20241220130020-removeTwitterFacebook-down.sql
ovbm: done

jstcki (approved):

PR #986, opened 2024-12-30
chore: remove calendar and matomo packages as well as daily statistics

ovbm:


hdahlheim (approved):

lucianak (changes requested):
Looks good, but some things are still missing:
- migrations need to be handled somehow, at the moment this would fail. Maybe we could introduce a new migration that deletes the calendar tables in a different module, and the js files in the migration module need to be deleted
will not break anything, but for cleanup: 
- remove calendar stuff from extractPersonalData script
- remove translations

ovbm:
> Looks good, but some things are still missing:
> 
> * migrations need to be handled somehow, at the moment this would fail. Maybe we could introduce a new migration that deletes the calendar tables in a different module, and the js files in the migration module need to be deleted
>   will not break anything, but for cleanup:
> * remove calendar stuff from extractPersonalData script
> * remove translations

Good point. What if we add a `sqls` folder in the backend-modules "migrations" and: 
1. copy and paste all the migration sqls files from the modules we want to delete into that folder
2. change the reference in the migration's javascript files

ovbm:
- Removed translation strings
- Removed reference in extractPersonalData
- Removed all migration.js files referencing calendar migrations

PR #987, opened 2024-12-30
chore: remove unused gsheets fields, updates, faq etc

ovbm:


apps/www/components/Marketing/Team.js
```diff
@@ -13,26 +13,27 @@ import SectionContainer from './Common/SectionContainer'
...
 const EmployeeLink = ({ employee, children }) => (
-  <Link href={`/~${employee.user.slug || employee.user.id}`} {...styles.link}>
+  <Link href={`/~${employee.userId}`} {...styles.link}>
     {children}
   </Link>
 )
 
 const Team = ({ employees, title, description }) => {
   const [colorScheme] = useColorContext()
+  const threeMarketingEmployees = employees.slice(1).sort(() => 0.5 - Math.random()).slice(0, 3)
```
jstcki, line 23: Hmm, this will probably lead to a hydration error because the set will be different on the server and client

apps/www/components/About/TeamTeaser.tsx
```diff
@@ -0,0 +1,31 @@
...
+        .then((data: { employees: EmployeeRecord[] }) =>
+          setEmployees(data.employees),
+        )
+    }
+  }, [employees])
+  if (!employees) {
+    return
+  }
+  // select twelve employees to show on about page
+  const selectedEmployees = employees
```
jstcki, line 21: Could move this filtering inside the useEffect?

jstcki (approved):
LGTM!

jstcki:
Another small thing: can we get rid of the headings on the team teaser? I don't think they make a lot of sense when we're just showing a random selection:

image

Also, the link to Impressum below the teaser opens in a new tab for some reason.

PR #993, opened 2025-01-09
chore: remove links to package ABO to force redirect to new shop

ovbm:
~~apps/www/components/Account/Memberships/Manage.js~~
~~Users with yearly abo and monthly abo are now directed to shop when they click on "Jahresmitgliedschaft kaufen" button~~ *needs to stay with ABO since we can't upsell new packages yet

apps/www/components/Frame/ProlongBox.js
The Prolongbox links to the new shop if the membership that is expiring can't be prolonged (yearly_abo for example) !membership.canProlong

apps/www/pages/teilen.js
The hard coded paynote on the teilen page now links to the new shop overview for both monthly and yearly

### DIFF
--- a/apps/www/components/Account/Memberships/Manage.js
+++ b/apps/www/components/Account/Memberships/Manage.js
@@ -97,7 +97,6 @@ const Actions = ({
                     <Link
                       href={{
                         pathname: '/angebote',
-                        query: { package: 'ABO' },
                       }}
                       passHref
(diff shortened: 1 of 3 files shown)


jstcki (approved):
YES – we could even link directly to `${process.env.NEXT_PUBLIC_SHOP_BASE_URL}/angebot/YEARLY` (or MONTHLY or whatever)

PR #996, opened 2025-01-15
PDF integration

annatraussnig:
Using print styles to generate a pdf document.

- [x] charts: show datawrapper png
- [x] infobox: show entire content
- [x] inline series nav: hide carousel & signup form
- [x] integration of the new pdf API (**needs testing**) 
- [x] remove image counter in pdf overlay (feels like an overkill to parse the mdast tree there)
- [x] fix (forced) darkmode (maybe too much effort for what's worth?)
- [x] match original image sizes (atm side by side images doesn't work)
- [x] fix republik logo covers content

### TODO:

- [x] end-to-end testing


packages/backend-modules/assets/express/pdf.js
```diff
@@ -17,15 +17,17 @@ module.exports = (server) => {
 
     // pick query parameters intended for PDF endpoint
     // (pass along rest as options, later)
-    const { images, download, size, ...options } = req.query
+    const { images, download, format, ...options } = req.query
 
     // build URL to fetch PDF from
+    // TODO: PDF_BASE_URL = https://screenshot.republik.ch/api/pdf
     const fetchUrl = new URL(`${PDF_BASE_URL}/${path}`)
```
jstcki, line 24: I _think_ this should probably just be `${PDF_BASE_URL}`, since the path is set on the `url` param?

packages/backend-modules/assets/express/pdf.js
```diff
@@ -17,15 +17,17 @@ module.exports = (server) => {
...
 
     // add params
+    fetchUrl.searchParams.set('url', `${FRONTEND_BASE_URL}/${path}`)
     images && fetchUrl.searchParams.set('images', images)
     download && fetchUrl.searchParams.set('download', download)
-    size && fetchUrl.searchParams.set('size', size)
+    format && fetchUrl.searchParams.set('format', format)
```
jstcki, line 30: We should also pass the `version` param: in the frontend we can add the publish date as version, e.g. `?version=2025-01-28T03%3A00Z`, so when an article is re-published, a new PDF will be generated

apps/www/components/Article/PdfOverlay.js
```diff
@@ -4,103 +4,75 @@ import {
...
 
-export const countImages = (element) => {
-  if (matchFigure(element) || matchVideo(element)) {
-    return 1
-  }
-  return (element.children || []).reduce(
-    (count, node) => count + countImages(node),
-    0,
-  )
+  pdfUrl.searchParams.set('url', articleUrl)
```
jstcki, line 27: I think this is needed:
```suggestion
  pdfUrl.searchParams.set('url', encodeURIComponent(articleUrl))
```

jstcki (approved):
1 tiny thing but let's gooooo

PR #998, opened 2025-01-16
elastic: update queries to proceed on conflict

lucianak:


hdahlheim (approved):
Let's hope this fixes it for now. In the long run we have to figure out how to reduce the dependency on Elastic Search for publishing.

PR #1000, opened 2025-01-20
Remove Pre-Commit Hooks

jstcki:
Removing the commitizen/commitlint/husky/lint-staged pre-commit hooks since we all agree that running `eslint --fix` when you want to commit has been 0 times useful and several times annoying, as it will always fail on some random old file.

Also the configuration seemed weird to me, as far as I understand: if husky is installed, we wouldn't need commitizen but I didn't look into it too much. I removed everything because it doesn't feel particularly useful to us anymore and if we need something like this in the future, we'll just add it back.

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,5 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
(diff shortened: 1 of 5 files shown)


hdahlheim:
good one for the 1000th PR 😂

PR #1004, opened 2025-01-23
fix: replace hard-coded usage of cdn.republik.space with cdn.repub.ch

jstcki:
cdn.republik.space is redundant and should be replaced with the main cdn URL cdn.repub.ch


hdahlheim (approved):

hdahlheim:
We could also extract the CDN base URL to a constant. Then again passing around the constant might lead to more errors so I'm okay with this change.

jstcki:
> We could also extract the CDN base URL to a constant. Then again passing around the constant might lead to more errors so I'm okay with this change.

Not sure it's worth it tbh. Most of it is either in references to files uploaded through Publikator, documentation examples or backend email templates, where we'd have to introduce Handlebars vars for this. In this case, time would probably better spent by improving the templating altogether.

hdahlheim (approved):

lucianak (approved):
Looks good

PR #1016, opened 2025-02-17
Verlegerinnen Survey

annatraussnig:
### Missing

- [x] Year field in personal details

SQL to migrate birthday to birthyear: 
`UPDATE users SET "birthyear" = EXTRACT(YEAR FROM birthday)::int WHERE birthday IS NOT NULL;`


packages/backend-modules/republik/migrations/sqls/20250218154508-user-birthyear-up.sql
```diff
@@ -0,0 +1,3 @@
+-- migrate up here: CREATE TABLE...
+ALTER TABLE "public"."users" ADD COLUMN "birthyear" integer;
+
```
hdahlheim, line 3: I would add the proposed birth year column migration as part of the up migration.

We did the same for the [profile link migration](https://github.com/republik/plattform/blob/27ae21bff3308552d14240dad07d4d646ca1da43/packages/backend-modules/auth/migrations/sqls/20241212154336-profileLink-up.sql).

hdahlheim (approved):
So I checked the backed again today. LGTM nothing to add :).

PR #1017, opened 2025-02-19
fix: wrap transactional prolong url in a link tag to avoid excessive line length

jstcki:


packages/backend-modules/mail/templates/membership_owner_autopay_failed.html
```diff
@@ -50,7 +50,7 @@
                       nicht belasten.
                     </p>
                     <p>
-                      Bitte aktualisieren Sie Ihre Kreditkartendaten {{prolong_url}} oder wählen 
+                      Bitte <a href="{{prolong_url}}">aktualisieren Sie Ihre Kreditkartendaten</a> oder wählen
```
lucianak, line 53: If I remember correctly we need to have links in a separate line in the html for mandrill to parse them correctly to the text version of the mail
```suggestion
                      Bitte 
                      <a href="{{prolong_url}}">aktualisieren Sie Ihre Kreditkartendaten</a>
                       oder wählen 
```

packages/backend-modules/mail/templates/membership_owner_autopay_failed.html
```diff
@@ -93,8 +93,8 @@
                       Ihre Mitgliedschaft am {{grace_end_date}} zu deaktivieren.
                     </p>
                     <p>
-                      So weit muss es nicht kommen: Begleichen Sie den ausstehenden 
-                      Jahresbeitrag {{prolong_url}} , wenn Sie möchten, auch mit einer 
+                      So weit muss es nicht kommen: <a href="{{prolong_url}}">Begleichen Sie den ausstehenden 
+                      Jahresbeitrag</a>, wenn Sie möchten, auch mit einer
```
lucianak, line 97: same here
```suggestion
                      So weit muss es nicht kommen: 
                      <a href="{{prolong_url}}">Begleichen Sie den ausstehenden 
                      Jahresbeitrag</a>
                      , wenn Sie möchten, auch mit einer 
```

lucianak (changes requested):
Looks good, thanks for fixing this :) ~~the links should just be in a separate line afaik (see comment)~~
edit: sorry, the links shouldn't have linebreaks

packages/backend-modules/mail/templates/membership_owner_autopay_failed.html
jstcki: Won't this add a space between the link and the comma?

PR #1019, opened 2025-02-20
Remove Matomo Pixel from Email Template

jstcki:


jstcki:
@hdahlheim @lucianak The statistics/matomo module is also still lingering around … feel free to also remove that if it's not called anymore.

lucianak (approved):
Thank you! 👍

hdahlheim (approved):

PR #1020, opened 2025-02-25
fix: bring back video crossorigin attribute to make subtitles work again

jstcki:


jstcki:
This doesn't work in Mobile Safari :(

PR #1022, opened 2025-02-27
fix: update firebase sdk

hdahlheim:
Attempt to fix the firebase push


hdahlheim:
- [x] Tested on love and android device

jstcki (approved):

PR #1023, opened 2025-02-27
fix: links to new shop via redirect instead of old ABO package for YEARLY_ABO

ovbm:


jstcki (commented):
We can also just link to `process.env.NEXT_PUBLIC_SHOP_BASE_URL`?

lucianak (approved):
thanks for fixing!

PR #1024, opened 2025-02-27
Feat/donations and reduced price in new shop

hdahlheim:


packages/backend-modules/payments/lib/services/CancelationService.ts
```diff
@@ -0,0 +1,116 @@
...
+  async revokeCancelation(sub: Subscription): Promise<Subscription> {
+    const tx = await this.db.transactionBegin()
+
+    try {
+      const cancelation = await tx.payments.subscriptionCancellations.find({
+        subscriptionId: sub.id,
+      })
+
+      if (cancelation) {
+        await tx.payments.subscriptionCancellations.delete(cancelation.id)
```
lucianak, line 80: would it make sense to keep a record of the cancellation but mark it as revoked?

packages/backend-modules/payments/lib/services/UserService.ts
```diff
@@ -0,0 +1,35 @@
...
+export class UserService {
+  protected pgdb: PgDb
+  protected users: UserDataRepo
+
+  constructor(pgdb: PgDb) {
+    this.pgdb = pgdb
+    this.users = new UserDataRepo(pgdb)
+  }
+
+  async updateUserName(
```
lucianak, line 14: where is this function used? It would be good to also update firstName and lastName on mailchimp if the name is updated

packages/backend-modules/payments/lib/config.ts
```diff
@@ -13,12 +13,15 @@ export type Config = {
...
   MONTHLY_SUBSCRIPTION_STRIPE_PRODUCT_ID: string
   YEARLY_SUBSCRIPTION_STRIPE_PRODUCT_ID: string
+  BENEFACTOR_SUBSCRIPTION_STRIPE_PRODUCT_ID: string
   REPUBLIK_STRIPE_SUBSCRIPTION_TAX_ID: string
   REPUBLIK_STRIPE_PAYMENTS_CONFIG_ID: string
   PROJECT_R_STRIPE_PAYMENTS_CONFIG_ID: string
   REPUBLIK_3_MONTH_GIFT_COUPON: string
   PROJECT_R_3_MONTH_GIFT_COUPON: string
   PROJECT_R_YEARLY_GIFT_COUPON: string
+  PROJECT_R_REDUCED_MEMBERSHIP_DISCOUNTS: string[]
```
lucianak, line 23: can you add the new env variables to example.env? :)

lucianak (approved):
Looks great! It's much more clean and logical now, love it! I only added a few small comments and have a few questions for understanding it better which are probably easier to just discuss :)

packages/backend-modules/payments/lib/services/CancelationService.ts
hdahlheim: Thought about that as well, for now I replicated the behaviour of the old setup. But there is at least one situation where checking the revoked cancellation would be useful.

packages/backend-modules/payments/lib/handlers/stripe/checkoutCompleted.ts
```diff
@@ -33,257 +43,392 @@ export async function processCheckoutCompleted(
...
+      } else {
+        console.log(`
+        No mailer configured
+
+        Thanks for the gift purchase
+
+        to: ${sess.customer_details.email}
+        gift-code: ${giftCodes[0]}
+        `)
+      }
```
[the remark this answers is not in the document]
hdahlheim: This to handle the case that the mailer was not passed to the Workflow class.

packages/backend-modules/payments/lib/config.ts
hdahlheim: will do

packages/backend-modules/payments/lib/services/CancelationService.ts
hdahlheim: Nevermind I missed the revoked field in the old table

PR #1026, opened 2025-03-04
strip query parameters from provided profileURL

ovbm:
- [ ] Test on Love

### DIFF
--- a/apps/www/components/Profile/EditProfile/Submit.js
+++ b/apps/www/components/Profile/EditProfile/Submit.js
@@ -27,6 +27,24 @@ const Submit = ({ me, user, t, state, setState, update }) => {
     return null
   }
 
+  // Function to ensure URL has protocol and strip parameters
+  const normalizeUrl = (url) => {
+    // Add protocol if missing
+    const urlWithProtocol = url.indexOf('https://') === -1 && 
+      url.indexOf('http://') === -1
+      ? 'https://' + url
+      : url
+    
+    // Strip parameters
+    try {
+      const urlObj = new URL(urlWithProtocol)
+      return `${urlObj.protocol}//${urlObj.host}${urlObj.pathname}`
+    } catch (e) {
+      // If URL parsing fails, return the URL with protocol
+      return urlWithProtocol
+    }
+  }
+
   const errorMessages = Object.keys(state.errors)
     .map((key) => state.errors[key])
     .filter(Boolean)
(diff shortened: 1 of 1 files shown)


ovbm:
Tested on love. Stripps query parameters reliably from provided profileUrl's

apps/www/components/Profile/EditProfile/Submit.js
```diff
@@ -27,6 +27,24 @@ const Submit = ({ me, user, t, state, setState, update }) => {
...
+    // Add protocol if missing
+    const urlWithProtocol = url.indexOf('https://') === -1 && 
+      url.indexOf('http://') === -1
+      ? 'https://' + url
+      : url
+    
+    // Strip parameters
+    try {
+      const urlObj = new URL(urlWithProtocol)
+      return `${urlObj.protocol}//${urlObj.host}${urlObj.pathname}`
```
annatraussnig, line 41: I feel like the purpose of this would be more transparent by setting the search/query params to '' and converting the url to string, but it's a detail.

apps/www/components/Profile/EditProfile/Submit.js
```diff
@@ -27,6 +27,24 @@ const Submit = ({ me, user, t, state, setState, update }) => {
     return null
   }
 
+  // Function to ensure URL has protocol and strip parameters
+  const normalizeUrl = (url) => {
+    // Add protocol if missing
+    const urlWithProtocol = url.indexOf('https://') === -1 && 
+      url.indexOf('http://') === -1
```
annatraussnig, line 34: Other detail: would use startsWith instead of indexOf

annatraussnig (approved):

apps/www/components/Profile/EditProfile/Submit.js
```diff
@@ -27,6 +27,24 @@ const Submit = ({ me, user, t, state, setState, update }) => {
...
+      ? 'https://' + url
+      : url
+    
+    // Strip parameters
+    try {
+      const urlObj = new URL(urlWithProtocol)
+      return `${urlObj.protocol}//${urlObj.host}${urlObj.pathname}`
+    } catch (e) {
+      // If URL parsing fails, return the URL with protocol
+      return urlWithProtocol
```
annatraussnig, line 44: Are we sure about this? Won't it lead to faulty urls in the DB? How about showing an error message instead?

ovbm: the test if it's a valid url happens in ProfileFieldsUrl.js before the submit. if it's not, the submit function will not even call.

ovbm:
@annatraussnig @lucianak 
Wouldn't it be better if this was done server-side? Or does it not matter much? Would love to hear your thoughts, so i can either close or merge this pr.

annatraussnig:
I don't see it as mattering much

lucianak:
I also think it doesn't matter that much, if we want to use this check to give direct feedback on the input it makes sense to have in the frontend imo

PR #1028, opened 2025-03-10
Article page cleanup

annatraussnig:
attempt to tame the article page beast some. by:

- extract low hanging fruits into own components
- extract identifier components to own file
- useSchema hook
- remove old paynotes/trynotes logic


apps/www/components/Article/DynamicComponents.tsx
```diff
@@ -0,0 +1,177 @@
+import { Loader } from '@project-r/styleguide'
+
+import dynamic from 'next/dynamic'
+
+// CAMPAIGN MODE
+// import { TrialPaynote } from '@app/app/(campaign)/components/trial-paynote'
```
jstcki, line 6: I guess we could also remove these campaign paynotes too

jstcki (approved):
Looks good!

Can we also already remove the main Article/Paynote components as well (not just the "mini" one)?

apps/www/components/Article/DynamicComponents.tsx
annatraussnig: ah yes!

PR #1032, opened 2025-03-12
refactor(www): unify paynote overlay visibility logic, hide from community page

jstcki:


apps/www/src/components/paynote-overlay/paynote-overlay.tsx
```diff
@@ -343,10 +343,31 @@ function PaynoteOverlayDialog() {
...
+): boolean {
   return (
+    (pathname === '/angebote' && searchParams.has('package')) ||
+    pathname === '/mitteilung' ||
+    pathname === '/anmelden' ||
+    pathname.startsWith('/konto') ||
+    pathname === '/meine-republik' ||
+    pathname === '/community' ||
+    searchParams.has('extract') ||
+    searchParams.has('extractId')
```
annatraussnig, line 358: would it make sense to have the list as an env variable? or at least a const?

jstcki: i thought about it but env doesn't make much sense because changing it would require a rebuild anyway. a const just adds more indirection because we differentiate between startsWith, query params, pathname …

PR #1033, opened 2025-03-17
Mark notifications as read

annatraussnig:
add button to mark all notifications as read (readers' request)

### DIFF
--- a/apps/www/components/Notifications/NotificationFeed.js
+++ b/apps/www/components/Notifications/NotificationFeed.js
@@ -1,3 +1,4 @@
+import { useMemo } from 'react'
 import {
   Interaction,
   Center,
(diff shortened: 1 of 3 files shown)


apps/www/components/Notifications/NotificationFeed.js
```diff
@@ -150,6 +172,17 @@ export default withT(
...
             </Link>
 
+            {!isEmpty && (
+              <button
+                {...linkStyleRule}
+                {...plainButtonRule}
+                style={{ marginLeft: 15 }}
+                onClick={() => markAllAsReadMutation().then(onReload)}
+              >
+                Alle als gelesen markieren
```
jstcki, line 182: Maybe use a translation string here?

jstcki (approved):

apps/www/components/Notifications/NotificationFeed.js
annatraussnig: excellent point

PR #1036, opened 2025-03-18
Regwall

annatraussnig:


apps/www/lib/context/MeContext.tsx
```diff
@@ -79,6 +86,30 @@ type MeContextValues = {
   hasAccess: boolean
   isEditor: boolean
   isClimateLabMember: boolean
+  trialStatus?: TrialStatusType
+}
+
+const getTrialStatus = (me?: MeObjectType | undefined): TrialStatusType => {
```
annatraussnig, line 92: this function could/should maybe come from the API? @lucianak

lucianak: so that would be my suggestion

PR #1040, opened 2025-03-20
Replace defaultProps with function parameters

jstcki:
React warns about defaultProps being deprecated on Function Components, so we replace them with default function parameters.

### DIFF
--- a/apps/www/components/Questionnaire/TextInput/TextInput.js
+++ b/apps/www/components/Questionnaire/TextInput/TextInput.js
@@ -38,7 +38,7 @@ const TextInput = (props) => {
   const [colorScheme] = useColorContext()
   const { t } = useTranslation()
 
-  const { label, onChange } = props
+  const { label = '', onChange } = props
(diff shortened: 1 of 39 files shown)


apps/www/components/Questionnaire/TextInput/TextInput.js
```diff
@@ -38,7 +38,7 @@ const TextInput = (props) => {
   const [colorScheme] = useColorContext()
   const { t } = useTranslation()
 
-  const { label, onChange } = props
+  const { label = '', onChange } = props
```
annatraussnig, line 41: don't we need the `text=''` too?

annatraussnig (approved):
Approved!

apps/www/components/Questionnaire/TextInput/TextInput.js
jstcki: That was already done on L33. I merged the two lines now :)

PR #1045, opened 2025-03-24
fix(SyntheticReadAloud): remove limit to one Pending derivative per repo

lucianak:


packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
```diff
@@ -246,40 +246,6 @@ export const derive = async (
   const { commitId } = getParsedDocumentId(document.id)
 
   if (user) {
```
annatraussnig, line 248: so the user is defined here…

packages/backend-modules/publikator/lib/Derivative/SyntheticReadAloud.ts
```diff
@@ -295,7 +261,7 @@ export const derive = async (
         status: 'Failure',
         result: { error },
         userId: user?.id,
```
annatraussnig, line 263: but we also check that the user is defined there. is there a reason for the double-check

annatraussnig (approved):
looks good :)

PR #1047, opened 2025-03-26
add download attribute to pgp key link

ovbm:


jstcki (commented):
Is this change even necessary at all? The API route already adds the Content-Disposition headers, so the file becomes downloadable (which is more reliable than the `download` attribute): https://github.com/republik/plattform/blob/29f0e54745a2515e5fc5864f0ab112898d751043/apps/www/pages/api/pgp/%5BuserSlug%5D.ts#L53-L57

ovbm:
Ah thanks, might not be necessary after all. I have an issue where closing the app while native download overlay is visible leads to the app opening up in the route of the file, bricking the app. So I have to do some more research into how to prevent the webview from restarting with a site with that header set.

ovbm:
solvied it: https://github.com/republik/republik-app/commit/82f0497e35201e841c18231ada6a2a8db2e772d3

PR #1051, opened 2025-04-03
fix(publikator): catch parse/serialize error when committing document

jstcki:
A quick and dirty fix for when a document cannot be parsed/serialized during a commit. The existing code would only catch GraphQL errors.

Instead of showing the spinner indefinitely, this is now displayed like other errors:

image

The real solution is to avoid transforming back-and-forth between Markdown and MDAST, like in #963


jstcki:
FYI @annatraussnig. We should finish and merge #963 at some point 😅

lucianak (approved):
Looks good!

PR #1052, opened 2025-04-08
feat(backend): log gql and timeout errors as single-line json

jstcki:
Logging errors as a single-line JSON is vastly better than logging multi-line messages to the console.

Mezmo can automatically parse this JSON, and will extract the `message` and `level` automatically.


hdahlheim:
Amazing! 

I agree we should look into a global solution as well. Maybe we can look into an adoption plan for a new logging solution after we are done with the current projects.

hdahlheim (approved):

lucianak (approved):
👍

PR #1059, opened 2025-04-30
remove current job add (hotfix)

ovbm:
removes the current job add https://republik.ch/2025/04/30/stellenausschreibung-trainee-unternehmensmanagement from paynotes.


ovbm:
tested

PR #1064, opened 2025-05-07
App users should still receive format and article notifications after their membership or trial runs out

lucianak:


ovbm (approved):
LGTM

PR #1068, opened 2025-05-08
prioritise trial status for paywall instead of metering

ovbm:
Currently the metering is checked before the trial status. Leading to an effect that on articles where reading is granted the paywall doesn't show up. This could be revered to show a more consistent experience, where after the trial ends the paywall always appears.


annatraussnig (approved):
yes we can do it like this too, if the other way was confusing people :)

PR #1073, opened 2025-05-14
remove audio buttons for non-members when they don't make sense

annatraussnig:
for non-members:

- remove play/audio queue buttons from feed
- remove audio queue button from article audio element (anonymous users don't have an audio queue)
- remove expand button from audio player (it expands into the queue manager, which doesn't make sense here)

bonus fix:

- remove clunky `canReadFreely` function and use better `isMember` attribute on `me`.


ovbm (approved):
LGTM

PR #1078, opened 2025-05-20
Paywall edits

annatraussnig:
- hide onboarding assistant for members without abo
- change copy of the paywall depending on the time since trial ended


lucianak (approved):
Fixed the if, now it seems to work :) I didn't test on iOS

PR #1080, opened 2025-05-21
chore: cleanup asset server package.json

hdahlheim:
- Removes all unused packages from the asset-server app
- Remove asset-server middleware from api
- Add assets package to packages that depend on it
- Move express files into the assets app


packages/backend-modules/documents/package.json
```diff
@@ -20,6 +20,7 @@
     "link": "yarn link"
   },
   "dependencies": {
+    "@orbiting/backend-modules-assets": "*",
```
jstcki, line 23: If I see correctly, this is only used to import 1 function that is only used by this module. This could easily be inlined, and this dependency removed

hdahlheim: Inlining would be an option but in this case I was wondering why we even convert webp to auto? the commit that introduced the `addFormatAuto` function does not give a reason: https://github.com/republik/plattform/commit/b128e29abcb6cdacc307b6481bcb98a2fbc71a62

jstcki: Yeah, I didn’t look too hard at what the code actally does😆

Probably in 2032 webp was not as well supported in our user base?

lucianak (approved):
LGTM ☀️

PR #1081, opened 2025-05-22
chore: remove pdf purge code for old pdf renderer

hdahlheim:
Removes KeyCDN purge code for pdfs from the publish mutation.

Since the introduction of https://github.com/republik/plattform/pull/996 PDF versions of Articles are no longer hosted on KeyCDN.


hdahlheim:
Closing in favour of removing the keyCDN module all together

PR #1099, opened 2025-06-04
Server Side Profile fetching

ovbm:
Previously, then navigating to ~slug, the server would fetch a users slug if that user had a public profile. That slug was then handed over to the profile component who would fetch userdata client side. This had the issue that if a search engine tried to index a site it would always receive an empty page even for users with public profiles. 

Since we now don't have to prevent links to articles to be displayed, we can query data server-side. The new implementation fetches the entire user profile server-side (not just a slug). This ensures that the page is SEO readable.

If no user is returned server side (private profile), the behaviour remains the same as before: a 404 page is shown. 

To enable the fetchMore function, an apollo client is initialized with the server-side fetched data as a cache, so that no duplicate request are made on page load.

In addition it adds structured data to help the public user sites being discovered and understood by search engines.

### DIFF
--- a/apps/www/components/Profile/index.tsx
+++ b/apps/www/components/Profile/index.tsx
@@ -1,6 +1,5 @@
-import { useQuery, useApolloClient } from '@apollo/client'
+import { useQuery } from '@apollo/client'
 import { css } from 'glamor'
-import { useEffect, useState } from 'react'
 
 import { useTranslation } from '../../lib/withT'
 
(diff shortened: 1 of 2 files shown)


jstcki (changes requested):
The manual cache initialization with client.writeQuery is luckily unnecessary. The key change to get SSR to work correctly was to just use the same `getPublicUser` query in `getServerSideProps` as on the profile page itself. Apollo Client will then use the cached data automatically.

There's no need for returning and checking for `publicUser`, setting the cache, and the skip/fetchPolicy options in `useQuery` (fetchPolicy is already cache-first).

PR #1100, opened 2025-06-04
Use materialised views for cockpit cache

hdahlheim:
Updates the resolvers used by the `/cockpit` page to fetch the data from materialised views instead of the Redis cache.

## WHY?
- Main benefit is that the data is no longer being evicted from the cache preventing the cockpit form breaking randomly.
- The views can also be used in our metabase queries.

## Testing
The current cockpit on staging fetches the data from materialised views already

## Not part of this PR:
- removal of the populate scripts for the redis cache
- removal of unused resolvers.

### DIFF
--- a/packages/backend-modules/discussions/graphql/resolvers/DiscussionsStats/evolution.js
+++ b/packages/backend-modules/discussions/graphql/resolvers/DiscussionsStats/evolution.js
@@ -20,6 +20,6 @@ module.exports = async (_, args, context) => {
 
   return {
     buckets: data.filter(({ key }) => keys.includes(key)),
-    updatedAt: data[0]?.updatedAt,
+    updatedAt: data[0]?.updatedAt || new Date(),
   }
 }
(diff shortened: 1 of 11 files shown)


packages/backend-modules/discussions/graphql/resolvers/DiscussionsStats/evolution.js
```diff
@@ -1,20 +1,10 @@
...
-    throw new Error(
-      'Unable to retrieve pre-populated data for DiscussionsStats.evolution',
-    )
-  }
-
-  // Retrieve pre-populated data.
-  const { result = [], updatedAt = new Date() } = data
+  const data = await context.pgdb.query(
+    'select * from cockpit_discussions_evolution;',
+  )
```
lucianak, line 7: for the other data you added a check that throws an error if the data is not there, I'd do that here as well :)

lucianak (approved):
Awesome! Thank you. I added some small comments

packages/backend-modules/discussions/graphql/resolvers/DiscussionsStats/evolution.js
hdahlheim: @lucianak I change the behaviour of the other resolvers to don't throw errors if there is no data.

PR #1102, opened 2025-06-05
create sitemap for articles

ovbm:
Creates two types of sitemaps:

- sitemap.xml -> an overview file of all yearly sitemaps
- /api/sitemap/[year] -> yearly sitemaps of all articles of that year

- /api/news-sitemap  -> overview files of all news-sitemaps
- /api/news-sitemap/[year] -> yearly sitemaps of all articles in the news sitemap format

I first tried to implement it in the app router with [sitemap.jt](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap ), since it's a special Route Handler that is cached by default. I think the current implementation uses an hourly cache which probably should be extended to a day.


ovbm:
Jeremy feedback: 

- [ ] to make it a bit simpler and consistent, use the sitemap file for the alternate sitemaps too, e.g. /news/sitemap.ts , [year]/sitemap.ts [year]/news/sitemap.ts : https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap#generating-multiple-sitemaps
- [ ] it's possible to group all these into src/app/(sitemaps)/  (the parentheses are invisible in the route path but keep the files in one place)
- [ ] put the gql queries into graphql/republik-api/queries/Sitemap.gql and use the generated queries with TS types

PR #1105, opened 2025-06-07
Fix profile url bug

ovbm:
require protocol for profile URL


hdahlheim:
This work for the profile update page but if someone updates the URLs using the API we still get corrupt values in the frontend, that can be executed. We need to add validation on the backend next.

ovbm:
I'm going to publish this and we do the api changes as a next step.

PR #1108, opened 2025-06-10
Dialog page cleanup

ovbm:
Screenshot 2025-06-12 at 17 48 20

What this PR does: 
- Introduces routes for dialog pages instead of rendering them via search params. Previously: `dialog?t=article&id=b19d55b3-d5f5-4df8-9e5f-da5475c94a3f` Now: `/dialog/2025/06/16/microsoft-es-knallt-in-den-kantonen`
- Adds redirect logic from old query to new dialog routes. 
- Removes all logic and references to "tab", the "t" in the previous dialog url, whose only job was to determine if it was the general dialog "Allgemeines Feedback" or an actual dialog page.
- Instead added a dedicated route /feedback for the Feedback dialog and linked it in the footer
- Refactored the /dialog landing page to feature the recent active dialogs more prominently. 
- Removes the focusID from canonical url's so that individual comments are not indexed as seperate sites or prevent the dialog from being indexed because search engines detect several pages with the exact same content but different canonical url's.


apps/www/components/ActionBar/utils.js
ovbm, line 1: I probably should clean up this file some more. It's such a mess...

jstcki (approved):

PR #1127, opened 2025-06-18
Fix Publikator rich-text copy paste

jstcki:
Fixes a super-weird build minification error in the `vfile-message` dependency of `unified`, which prevented pasting rich text content. The only fix that worked was to update unified and related packages to their latest versions in Publikator.


apps/publikator/components/editor/modules/document/createPasteHtml.js
```diff
@@ -28,45 +28,54 @@ const PARAGRAPH_TYPES = ['PARAGRAPH', 'INFOP', 'BLOCKQUOTEPARAGRAPH']
   TEXT
 } = TRANSFER_TYPES*/
 
-export default (centerModule, figureModule) => (event, change, editor) => {
-  const transfer = getEventTransfer(event)
-  const cursor = editor.value.selection.anchorKey
-  const blockType = editor.value.document.getClosestBlock(cursor).type
+const createPasteHtml =
```
jstcki, line 31: FYI, most of the lines below are formatting changes because ESLint didn't like that we're exporting an anonymous function.

PR #1129, opened 2025-06-18
html escape comments before saving

hdahlheim:


packages/backend-modules/discussions/graphql/resolvers/_mutations/submitComment.js
```diff
@@ -79,7 +80,7 @@ module.exports = async (_, args, context) => {
       discussionId: discussion.id,
       parentId,
       userId: me.id,
-      content,
+      content: DOMPurify.sanitize(content),
```
jstcki, line 83: Just to be clear: this still allows HTML but is supposed to [strip bad stuff](https://github.com/cure53/DOMPurify?tab=readme-ov-file#what-does-it-do). Another option would be to disallow any HTML by setting `ALLOWED_TAGS` to `[]`.

jstcki (commented):
Should we also do this in other places like the user profile?

packages/backend-modules/discussions/graphql/resolvers/_mutations/submitComment.js
hdahlheim: Yes I agree we should disallow html because the things that are allowed should be covered by the markdown transformation. I actually not sure if we should escape the string instead of stripping the tags but that could lead to breaking the markdown quote syntax.

PR #1143, opened 2025-06-26
Fix mediaProgress errors

jstcki:
This resulted in `upsertMediaProgress` having been called for logged out users

Also catch errors in debounced media progress mutations which would have been reported as unhandled promise rejections

### DIFF
--- a/apps/www/components/Audio/AudioPlayerController.tsx
+++ b/apps/www/components/Audio/AudioPlayerController.tsx
@@ -164,12 +164,7 @@ const AudioPlayerController = ({ children }: AudioPlayerContainerProps) => {
 
   const handleError = (error: Error | string) => {
     setHasError(true)
-    if (typeof error === 'string') {
-      reportError('handle audio-error', error)
-    } else {
-      error = JSON.stringify(error, Object.getOwnPropertyNames(error), 2)
-      reportError('handle audio-error', error)
-    }
+    reportError('handle audio-error', error)
     trackEvent([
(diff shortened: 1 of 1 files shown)


apps/www/components/Audio/AudioPlayerController.tsx
```diff
@@ -164,12 +164,7 @@ const AudioPlayerController = ({ children }: AudioPlayerContainerProps) => {
 
   const handleError = (error: Error | string) => {
     setHasError(true)
-    if (typeof error === 'string') {
-      reportError('handle audio-error', error)
-    } else {
-      error = JSON.stringify(error, Object.getOwnPropertyNames(error), 2)
```
jstcki, line 170: This created a lot of weird-looking reports in Sentry and is completely unnecessary.

apps/www/lib/context/MeContext.tsx
```diff
@@ -216,7 +216,8 @@ const MeContextProvider = ({ children, assumeAccess = false }: Props) => {
         isEditor: checkRoles(me, ['editor']),
         isClimateLabMember,
         trialStatus,
-        progressConsent: me?.progressOptOut !== true,
+        // Progress consent requires a logged in user AND them not having opted out
+        progressConsent: me && me.progressOptOut !== true,
```
ovbm, line 220: This is equivalent, no? I mean yours is probably more readable but it should do the same.

ovbm (approved):

PR #1146, opened 2025-07-02
hide currently read bookmark

annatraussnig:
A few small changes:

1. we query 6 bookmarks, but only show five. If the one bookmark is the article currently read, if hide the corresponding bookmark
2. we reduce the padding of the most read feed and fix an issue with padding in mobile
3. we add a 6th article in the most commented feed on desktop and increase padding when there is no image
4. we make sure that all bookmarks are documents


jstcki:
A slightly unrelated thing I noticed when checking out the preview: when there are no documents returned in a recommendation query, it will stay in loading limbo forever:
image

jstcki (approved):
👍 👍 👍

PR #1149, opened 2025-07-02
Audio Player Bugfixes

jstcki:
Just fixing a few things that popped up in Sentry


jstcki:
This actually fixes #811 😂

PR #1155, opened 2025-07-08
remove email plausible tracking

ovbm:
1. Doesn't inject the image into the mailchimp html on publishing a newsletter
2. Only returns the image o the email-open route, without sending the request to plausible. We do this because if we return nothing, email clients will complain that the email content couldn't be loaded properly. We can remove this route in the future, when it's unlikely that older emails are being openend anymore.


jstcki (approved):
👍

PR #1156, opened 2025-07-09
Fix Comment Submissions

jstcki:
In some cases, updating the cache after a comment submission doesn't work because the cache is empty. Then, we just do nothing.

Also updates to the more recent [`cache.updateQuery`](https://www.apollographql.com/docs/react/caching/cache-interaction#using-updatequery-and-updatefragment) instead of using `readQuery` and `writeQuery`.


jstcki:
There's still the slight issue where the GraphQL subscription will tell you in a green badge that there is a new comment before the mutation actually is returned and the cache is updated but I think that's okay for now.

https://github.com/user-attachments/assets/819a435f-7468-48b4-b1b3-cb78a85ffa36

PR #1158, opened 2025-07-10
feat: add slack notifier for subscription cancellations

hdahlheim:
a feature we forgot back when we introduced the new subscriptions

### DIFF
--- a/packages/backend-modules/payments/graphql/resolvers/_mutations/cancelMagazineSubscription.ts
+++ b/packages/backend-modules/payments/graphql/resolvers/_mutations/cancelMagazineSubscription.ts
@@ -2,10 +2,7 @@ import { GraphqlContext, User } from '@orbiting/backend-modules-types'
 import { default as Auth } from '@orbiting/backend-modules-auth'
 import { Subscription } from '../../../lib/types'
 import { PaymentService } from '../../../lib/services/PaymentService'
-import {
-  CancallationSlackNotifier,
-  CancellationService,
-} from '../../../lib/services/CancellationService'
+import { CancellationService } from '../../../lib/services/CancellationService'
 import { SubscriptionService } from '../../../lib/services/SubscriptionService'
 
 type CancellationInput = {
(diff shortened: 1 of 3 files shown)


packages/backend-modules/payments/lib/services/CancellationService.ts
```diff
@@ -63,7 +97,7 @@ export class CancellationService {
           },
         )
 
-    return await this.billingRepo.updateSubscription(
+    const newLocalSub = await this.billingRepo.updateSubscription(
```
lucianak, line 100: sorry to be nitpicky, I find the name a bit confusing. Maybe updatedSub or updatedSubscription?

packages/backend-modules/payments/graphql/resolvers/_mutations/cancelMagazineSubscription.ts
```diff
@@ -30,19 +33,24 @@ export = async function cancelMagazineSubscription(
...
+  const owner = await getSubscriptionOwner(ctx, sub)
+  if (!owner) {
+    throw Error('api/unexpected')
+  }
 
-  const cs = new CancellationService(new PaymentService(), ctx.pgdb)
+  Auth.Roles.ensureUserIsMeOrInRoles(owner, ctx.user, ['admin', 'supporter'])
+
+  const cs = new CancellationService(new PaymentService(), ctx.pgdb, [
+    new CancallationSlackNotifier(),
```
lucianak, line 44: also nitpicky: would it make sense here to instantiate the CancallationSlackNotifier inside the constructor of the CancellationService instead of passing it? Like that you wouldn't have to pass it as argument every time and I feel like it'd be more readable

lucianak (approved):
looks great, I like it! I just made some small nitpicky suggestions

packages/backend-modules/payments/lib/services/CancellationService.ts
hdahlheim: I agree, my imagination was a bit limited yesterday 😄

PR #1161, opened 2025-07-15
remove links to discussions from articles of closed

ovbm:
DiscussionLinks are displayed in the ActionBar based on if a discussionId is returned from the utils function. We simplified this logic significantly with this PR: https://github.com/republik/plattform/pull/1108/files#diff-86982bc6e1fa863a7dca70981ae6a6a571aeb154130da36b7f93cbc45f110100 but missed that "closed" discussions would return undefined discussionId and thus prevent rendering of the Links.

Tested locally againts the live API

### DIFF
--- a/apps/www/components/ActionBar/DiscussionLinkButton.js
+++ b/apps/www/components/ActionBar/DiscussionLinkButton.js
@@ -23,7 +23,7 @@ const DiscussionLinkButton = ({
   const getLabel = () => {
     if (useCallToActionLabel) {
       return t.pluralize('article/actionbar/discussion/call-to-action', {
-        count: discussionCount || 0,
+        count: discussionCount,
       })
     }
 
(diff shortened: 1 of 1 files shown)


apps/www/components/ActionBar/utils.js
```diff
@@ -5,11 +5,16 @@ export const getDiscussionLinkProps = (
...
 
-  const discussionPath =
-    template === 'discussion' ? discussion?.path : `/dialog${discussion?.path}`
   const isDiscussionPage = template === 'discussion'
+
+  const discussionPath = isDiscussionPage
+    ? discussion?.path
+    : `/dialog${discussion?.path}`
+
   const discussionCount = discussion?.comments?.totalCount
```
jstcki, line 14:
```suggestion
  const discussionCount = discussion?.comments?.totalCount ?? 0
```

jstcki (approved):

PR #1163, opened 2025-07-17
Fix AudioQueue Overfetching

jstcki:
Progress is already queried in the initial audio queue query, so we DON'T NEED TO DO IT AGAIN FOR EACH INDIVIDUAL QUEUE ENTRY!


ovbm (approved):
Looks good.

PR #1166, opened 2025-07-18
add pino logger

hdahlheim:
Our homegrown logger setup hasn't aged well. This PR introduces the pino logger to our backend.

By using the pino-http plugin we can make sure that all logs that happen in the HTTP layer are connected to a request id.
In production pino generates structured json logs that reduce the clutter in our log drain.

- [x]: added pino-http to all http requests
- [x]: inject pino into the pgboss queue

TODO: refactor console and debug calls to pino

### DIFF
--- a/packages/backend-modules/base/express/graphql.js
+++ b/packages/backend-modules/base/express/graphql.js
@@ -2,6 +2,7 @@ const { ApolloServer } = require('apollo-server-express')
 const { makeExecutableSchema } = require('@graphql-tools/schema')
 const { SubscriptionServer } = require('subscriptions-transport-ws')
 const { execute, subscribe } = require('graphql')
+const { logger } = require('@orbiting/backend-modules-logger')
 
 const cookie = require('cookie')
 const cookieParser = require('cookie-parser')
@@ -38,6 +39,7 @@ module.exports = async (
       req,
       scope,
       documentApiKey,
+      logger: req?.log || logger,
       user: global && global.testUser !== undefined ? global.testUser : user,
     })
     // prime User dataloader with me
--- a/packages/backend-modules/base/express/graphql.js
+++ b/packages/backend-modules/base/express/graphql.js
@@ -125,14 +125,27 @@ module.exports = async (
         async requestDidStart() {
           return {
             async didEncounterErrors({ context, request, errors }) {
-              console.error(
-                JSON.stringify({
-                  req: context.req._log(),
-                  message: `GraphQL error for operation '${request.operationName}'`,
-                  level: 'ERROR',
-                  graphQLErrors: errors,
-                }),
-              )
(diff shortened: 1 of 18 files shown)


hdahlheim:
@jstcki @lucianak If you both could take a look and give feedback. I don't think refactoring everything as part of this PR makes sense. The payments worker folder contains most of the "examples" on how the new logging can be used.

jstcki:
I agree with not refactoring everything here but changing `req._log` to `req.log` seems pretty straightforward, so maybe still worth it?

hdahlheim:
@jstcki @lucianak, I removed all calls to `_log()` and replaced them with the appropriate pino log call. The PR got much bigger because of it, sorry for that, but could you please review it again?

jstcki (approved):
LGTM :)

jstcki:
(I haven't actually checked the logs to verify that everything works as intended)

packages/backend-modules/auth/lib/Users.js
```diff
@@ -307,7 +307,7 @@ const signIn = async (
 }
 
 const shouldAutoLogin = ({ email }) => {
-  if (AUTO_LOGIN_REGEX) {
+  if (process.env.NODE_ENV !== 'production' && AUTO_LOGIN_REGEX) {
```
lucianak, line 310: 💖

packages/backend-modules/base/express/graphql.js
```diff
@@ -123,14 +125,27 @@ module.exports = async (
...
-                  graphQLErrors: errors,
-                }),
-              )
+              if (context.logger.isLevelEnabled('debug')) {
+                context.logger.debug(
+                  {
+                    graphqlRequest: {
+                      query: request.query,
+                      variables: request.variables,
+                      errors: errors,
```
lucianak, line 134: would it be possible to log the errors using logger.error also when debug is enabled? I feel like that would make filtering logs easier. But not sure if that's feasible

packages/backend-modules/payments/lib/workers/StripeWebhookWorker.ts
```diff
@@ -95,22 +107,27 @@ export class StripeWebhookWorker extends BaseWorker<WorkerArgsV1> {
...
-      console.error(e)
+
       throw e
     }
 
-    console.log(
-      'successfully processed stripe event %s [%s]',
-      event.id,
-      event.type,
+    this.logger.info(
```
lucianak, line 128: why is this info and not debug as for the other workers?

lucianak (approved):
looks good!

packages/backend-modules/payments/lib/workers/StripeWebhookWorker.ts
hdahlheim: To be able to see the health of stripe event processing

PR #1172, opened 2025-07-23
Optimize Yearly Overview

jstcki:
Rendering a yearly overview of (e.g. /2023) would request a screenshot for each Front teaser of an archived year. In two sizes. Each of these screenshot requests would trigger a query to the backend to only extract and render a single teaser. You do the math.

This PR changes this, so each year's Front is prerendered statically as one page. The screenshot service then only makes requests to this static page and extracts the teaser based on its DOM element ID.

So 100s of requests to the backend for a single visitor to this page become 0.


annatraussnig:
should we close this one?

PR #1178, opened 2025-08-13
Feat/u30 coupon create script

hdahlheim:


hdahlheim:
@jstcki the coupon name should now be correct right?

jstcki (approved):

PR #1200, opened 2025-08-29
chore: allow node engine 22

hdahlheim:


jstcki:
We should probably do this if no blockers exist? Maybe go straight to 24, which is already the active release?

PR #1203, opened 2025-09-03
Properly calculate donations for subscriptions/shop in calculateKpi

lucianak:
- [x] include benefactor subscriptions in calculateKpis
- [x] include shop donations in calculateKpis
- [x] include donations by paying more for subscriptions in calculateKpis
- [x] save payment intent with orders and charges 
- [x] always save charge on charge.succeeded
- [x] test calculateKpi changes
- [x] test code changes on love
- [x] modify calculateKpis to include info from single donation charges
- [ ] backfill missing charges
- [x] add database view for KPI query

### DIFF
--- a/packages/backend-modules/republik/script/finance/calculateKpis.js
+++ b/packages/backend-modules/republik/script/finance/calculateKpis.js
@@ -56,120 +56,45 @@ const evaluateCompanyMonth = async (
   pgdb,
 ) => {
   const query = `
-    SELECT
-      pay.id,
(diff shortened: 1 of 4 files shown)


packages/backend-modules/republik/script/finance/calculateKpis.js
hdahlheim, line 1: Would it be beneficial to extract the queries in this file into a set of database views? They don't need to be materialised but it would allow us to also display them in Metabase.

PR #1204, opened 2025-09-04
remove userProgress from default document queries add it's own

ovbm:
This PR fixes an age old problem, where the Inline Series Navigation would flicker when users scrolled past. This was occurring because the entire document was passed through the Series Navigation to "feed" it's ActionBar with the userProgress fields on that document (never mind that the ActionBar wasn't even rendered in the inline Series Navigation). 

It does so by removing the userProgress fields from all document queries and extending the useProgress API to be able to return a document progress for a specific document. That way, when upsertDocumentprogress is called, the document doesn't change, but the respective getDocumentProgress query is called for the document. 

This also means however that the feed for example makes one additional graphql request for each of it's FeedItems when it loads. Is this a problem? Is the removal of the userProgress from the default queries not even necessary, and are the changes in the actionbar and Progress components enough?


apps/www/graphql/republik-api/queries/DocumentProgressQuery.gql
```diff
@@ -0,0 +1,14 @@
+query getDocumentProgress($path: String!) {
+  document(path: $path) {
+    id
+    userProgress {
+      percentage
```
jstcki, line 5: Nit: Are we using this value anywhere? If not, maybe we don't need to fetch it.

jstcki (changes requested):
Very nice! I think it makes a lot of sense to remove the dynamic part from the article/document query.

If we display it as read-only value in the feed, we should keep it in the document query though. It's definitely not a great idea to fetch the progress for each feed entry individually. This creates another bottleneck like the one we recently fixed on the audio playlist.

I think what's needed here is a different <UserProgress> component on the article and in the feed/bookmark action bars. The one on the feed should just take a percentage as prop and display it.

And another idea for future work building on this: there seem to be other dynamic values that are queried on the document like mediaProgress, unreadNotifications and userBookmark which *could* create similar problems. We should also extract these from the document query and just load them in the place where they are displayed.

apps/www/graphql/republik-api/queries/DocumentProgressQuery.gql
ovbm: sooo... it doens't upsert document progress when the current node's percentage is the same as the userprogress??? might be a mistake

jstcki (approved):

PR #1211, opened 2025-09-15
remove negative margin rule in paynote

annatraussnig:
fixes a bug causing the overlay to overflow the body of the page.


annatraussnig:
the removal of the rule doesn't seem to cause any unwanted side effect, but feel free to have a thorough look later, @jstcki.

PR #1222, opened 2025-10-01
apollo server v5 update

hdahlheim:
Updates apollo server to version 5 from version 3

### DIFF
--- a/packages/backend-modules/base/express/graphql.js
+++ b/packages/backend-modules/base/express/graphql.js
@@ -1,5 +1,12 @@
-const { ApolloServer } = require('apollo-server-express')
-const { ApolloServerPluginDrainHttpServer } = require('apollo-server-core')
+const { ApolloServer } = require('@apollo/server')
+const {
+  ApolloServerPluginDrainHttpServer,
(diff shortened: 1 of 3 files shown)


packages/backend-modules/base/express/graphql.js
```diff
@@ -115,27 +118,20 @@ module.exports = async (
...
-      ) {
-        return JSON.parse(JSON.stringify(response).replace(/\u2028/g, ''))
-      }
-      return response
-    },
     plugins: [
+      ApolloServerPluginLandingPageProductionDefault({
+        footer: false,
+      }),
       // https://www.apollographql.com/docs/apollo-server/v3/api/plugin/drain-http-server
```
jstcki, line 125:
```suggestion
      // https://www.apollographql.com/docs/apollo-server/api/plugin/drain-http-server
```

jstcki (commented):
Looks good to me!

hdahlheim:
> Looks good to me! Not super urgent, but could you also check (if you haven't done so yet) that our Apollo Studio config still works, and that our integration test setup is still okay?

I checked all of that. Test work and Apollo studio also works fine.

PR #1227, opened 2025-10-03
replace copy-paste pipeline with custom html-to-slate serializer

annatraussnig:
This changes adds a fair bit code, but it should make the copy pasting more deterministic.


ovbm (approved):
LGMT

PR #1393, opened 2026-04-20
add format to meta

annatraussnig:
resolve format and contributors metadata and send to tts server


hdahlheim (changes requested):
Look alright to me. Could you maybe remove the IA like comments, I think the clutter the code and might become out of date rather quickly even though this code will probably not exist for much longer once we migrate CMS.

PR #1403, opened 2026-05-19
Next 16

jstcki:
Upgrade all the apps to Next.js v16 for compatibility with the latest next-sanity package (required by #1402)

Caveat: this breaks www comment links, but on Vercel _only_, where path matching is broken (the `x-matched-path` header matches on the top-level `[...path]` route instead of `/dialog/[...path]`). After confirming that this does not happen on Heroku, we accept this issue in order to move forward.


apps/admin/next.config.js
```diff
@@ -1,7 +1,3 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
-const withBundleAnalyzer = require('@next/bundle-analyzer')({
-  enabled: process.env.ANALYZE === 'true',
-})
```
annatraussnig, line 4: not using this anymore?

apps/admin/proxy.ts
```diff
@@ -1,6 +1,6 @@
 import { NextRequest, NextResponse } from 'next/server'
 
-export async function middleware(req: NextRequest) {
+export async function proxy(req: NextRequest) {
```
annatraussnig, line 3: just WHY, next????

apps/www/src/components/Discussion/shared/CommentLink.js
```diff
@@ -1,40 +1,21 @@
...
-    }
-  }
+  const pathPrefix =
+    discussion.document?.meta?.template === 'article' ? '/dialog' : ''
+
+  const href = comment?.id
+    ? `${pathPrefix}${discussion.path}?focus=${comment.id}`
+    : `${pathPrefix}${discussion.path}`
+
+  return href
```
annatraussnig: are all these dialog changes in this pr's scope?

apps/www/package.json
```diff
@@ -10,12 +10,12 @@
...
     "dev": "concurrently \"PORT=3010 next dev\" \"yarn gql:types --watch\" \"panda --watch\"",
     "start": "next start",
-    "build": "panda && yarn gql:types && next build",
+    "build": "panda && yarn gql:types && next build --webpack",
     "test": "jest",
     "test:watch": "jest --watch",
     "cloc": "find . -name '*.js' -not -path \"./node_modules*\" -not -path \"./.next*\" | xargs wc -l",
     "lint": "next lint",
-    "analyze": "ANALYZE=true next build",
+    "analyze": "next experimental-analyze",
```
annatraussnig, line 18: ah, this answers my first question

annatraussnig (approved):

apps/admin/next.config.js
jstcki: next 16 has a built-in one (did you ever use this?)

apps/www/src/components/Discussion/shared/CommentLink.js
jstcki: yes/no, they were attempts to fix the broken vercel behavior but also don't make things worse since we were getting deprecation warnings on using the 'url' package anyway.

PR #1408, opened 2026-05-28
Error reporting for UTM session storage

ovbm:
Catches the error when trying to set or read utm session storage via sentry. Could help us understand utm storage flows better.


jstcki (approved):
This probably will only show us SecurityErrors (e.g. when a user disallows cookies) but let's see if this will be useful. https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage#exceptions

PR #1411, opened 2026-06-02
Add browserslist config to www for Safari 15 support

jstcki:
Next.js 16 supports Safari 16.4+ by default, adding explicit Safari 15.4 support should fix JavaScript not running.

See https://nextjs.org/docs/architecture/supported-browsers


ovbm (commented):
I think we should use `ios_saf`, because according to browserlist that's how you target iOS browsers specifically. `safari` is used for desktop. See: https://github.com/browserslist/browserslist#browsers

jstcki:
Tested on iOS 15.8 ✅

PR #1416, opened 2026-06-22
remove podcast buttons from android

ovbm:
Google has rejected our appeal that linking to Spotify doesn't violate their terms of services, so I have to hide all Podcast Buttons from android apps: ActionBar and end of article page.


ovbm:
Two appeals failed. Merging and deplying.

PR #1417, opened 2026-06-24
show device info

ovbm:
Uses the device table for session view in admin tool instead of the session user agent, which is only set once:

Screenshot 2026-06-24 at 11 39 41

### DIFF
--- a/apps/admin/components/Zat/User/Sessions.js
+++ b/apps/admin/components/Zat/User/Sessions.js
@@ -7,6 +7,15 @@ export const fragments = gql`
     sessions {
       id
       userAgent
+      device {
+        information {
+          appVersion
+          os
+          osVersion
+          model
+        }
+        lastSeen
+      }
     }
   }
 `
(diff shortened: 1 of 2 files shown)


apps/admin/components/Zat/User/Sessions.js
```diff
@@ -7,15 +7,28 @@ export const fragments = gql`
...
 export const Sessions = ({ sessions }) =>
   !!sessions?.length && (
     <div {...styles.part}>
-      {sessions.map(({ id, userAgent }) => (
-        <div key={id}>{userAgent}</div>
+      {sessions.map(({ id, userAgent, device }) => (
+        <div key={id}>
+          {device
+            ? `${device.information.os.toUpperCase()} ${device.information.osVersion} · ${device.information.model} · App ${device.information.appVersion}`
+            : userAgent}
```
jstcki, line 30: maybe add a note that the session UA info might be outdated?

jstcki (approved):

apps/admin/components/Zat/User/Sessions.js
ovbm: good idea

ovbm: Screenshot 2026-06-24 at 14 21 57
done

PR #1428, opened 2026-07-21
Share image

ovbm:
Replicates the share image generation for sanity articles with a dedicated nextjs og route


apps/www/src/app/(sanity)/api/og/route.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { ImageResponse } from 'next/og'
+import type { NextRequest } from 'next/server'
+import { loadShareImageFonts } from '../../components/share-image/fonts'
+import {
+  SHARE_IMAGE_HEIGHT,
+  SHARE_IMAGE_WIDTH,
+  ShareImage,
+} from '../../components/share-image/share-image'
+
+export const runtime = 'edge'
```
jstcki: Is this an improvement over the default (node)?

apps/www/src/app/(sanity)/api/og/route.tsx
```diff
@@ -0,0 +1,79 @@
...
+      height: SHARE_IMAGE_HEIGHT,
+      fonts: fonts.map((font) => ({
+        name: font.name,
+        data: font.data,
+        weight: font.weight,
+        style: font.style,
+      })),
+      headers: {
+        'cache-control':
+          'public, immutable, no-transform, max-age=300, s-maxage=3600, stale-while-revalidate=86400',
```
jstcki: I think we should disable the cache headers in dev

jstcki (changes requested):
Looks good. One omission: this should also be added to `pages/[...path]/page.tsx` too, since we use the same SEO data as on articles for those.

apps/www/src/app/(sanity)/api/og/route.tsx
ovbm: No, it was recommended for next < v13, but not anymore. removed it

apps/www/src/app/(sanity)/api/og/route.tsx
ovbm: done

jstcki (approved):

PR #1433, opened 2026-07-23
Sanity darkmode clean

ovbm:
Changes the way theme.darkMode is handled. Not via useEffect but the page/article renders a `data-force-theme="dark"` marker on its content wrapper, and the Panda dark condition matches it from the root via `:root:has([data-force-theme="dark"])`.  `:has()` is anchored on , so it matches regardless of how deep the marker sits.
The sanity preview toggle is handled via useEffect and theme-provider

Tested and both the preview toggle and darkmode theme selection create the desired effect. darkmode theme requires refresh of the preview (expected), preview toggle is instant

### DIFF
--- a/apps/www/src/app/(sanity)/components/preview-theme-listener.tsx
+++ b/apps/www/src/app/(sanity)/components/preview-theme-listener.tsx
@@ -5,10 +5,8 @@ import { useSetManualTheme } from '@/app/components/theme-provider'
 
 /**
  * Receives the dark-mode toggle from the Sanity Studio preview header and turns
- * it into a manual `forcedTheme` override. The Studio and this app run on
- * different origins, so the toggle can only reach us via `postMessage` (see the
- * PreviewHeader component in the studio repo). The message shape is the shared
- * contract between the two repos; `theme` is `'dark'` or `undefined` (off).
+ * it into a manual `forcedTheme` override. The message shape is the shared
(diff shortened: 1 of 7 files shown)


jstcki (approved):
Checked that it works as intended. I think it's suboptimal that we have a <Theme> component which injects a style as well as attributes on article/page level. Preferably this logic would be centralized but we can clean up later. :shipit: 

Note: There is no way to force light mode when the system/user settings are set to dark but I think that's an omission in https://github.com/republik/studio/pull/71, will comment there.

apps/www/src/app/(sanity)/components/preview-theme-listener.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { useEffect } from 'react'
+import { useSetManualTheme } from '@/app/components/theme-provider'
+
+/**
+ * Receives the dark-mode toggle from the Sanity Studio preview header and turns
+ * it into a manual `forcedTheme` override. The Studio and this app run on
+ * different origins, so the toggle can only reach us via `postMessage` (see the
+ * PreviewHeader component in the studio repo). The message shape is the shared
+ * contract between the two repos; `theme` is `'dark'` or `undefined` (off).
```
jstcki: This comment is/will be off: 'light' should also be possible

PR #1437, opened 2026-07-28
Fix article dynamic components in the app router (next/router → next/navigation)

ovbm:
## Problem

The 24 components in `Article/DynamicComponents.tsx` are now rendered by the new app-router article page too, via `LegacyDynamicComponent`. **9 of the 24 crashed there.**

Root cause is one mechanism: `next/router`'s `useRouter()` throws `NextRouter was not mounted` when there's no `RouterContext` (`next/dist/client/router.js:143-151`)This is the same failure that was already patched in `ChallengeAcceptedInlineTeaser`.

## Changed components

**Crashed on first render (6):**

| Identifier | Files | Change |
|---|---|---|
| `TESTIMONIAL_LIST` | `Testimonial/List.js` | `shallow` dropped, so both calls are plain `router.replace/push('/community')` |
| `ELECTION_CANDIDACY` | `Vote/ElectionCandidacy.js` | Class component, so it can't use hooks. It only needed `?edit` once at construction, so that single value is injected as `initialIsEditing` and the class no longer touches a router at all |
| `QUESTIONNAIRE` | `Questionnaire/Questionnaire.js` | Object href → string (`next/navigation` only accepts strings) |
| `QUESTIONNAIRE_SUBMISSIONS` | `Submissions/legacy/index.js` | Reads via `useSearchParams`; the search-sync effect now reads its own derived `searchQuery`/`sortBy`/`sortDirection` instead of re-reading `router.query` |
| `QUESTIONNAIRE_OVERVIEW` | `Submissions/index.tsx`, `components/Links.tsx`, `QaBlock.tsx`, `QuestionnaireMeta.tsx`, `ShareImage.tsx`, `views/SubmissionView.tsx` | Query params → `useSearchParams`. The local `getPathname(router)` helper was deleted — it was literally `usePathname()` |
| `EDGE_QUESTION` | `Climatelab/EdgeQuestion/index.tsx`, `QuestionScroll.tsx` | Query params, and `asPath` → `usePathname()` + search |

**Rendered fine, crashed one step in (3).** These are the ones worth reviewing carefully — a smoke test would have missed them:

| Identifier | Trigger | Fixed in |
|---|---|---|
| `TRIAL_FORM` | **User submits their email** — `LoginForm` swaps to `CodeForm` | `auth/login/code-form.tsx` |
| `POSTCARD` | Logged-in user **with the `climate` role** → `PostcardGenerator` → `QuestionnaireWithData` | fixed via `Questionnaire.js` |
| `INSTANT_SURVEY` | Survey slug 404s → `<StatusError statusCode={404}>` | `StatusError/index.js` |

`TRIAL_FORM` is the most user-visible: it's a regwall CTA that failed at step 2 of signup, so the trial funnel died with a blank error and no obvious cause.

**Unchanged and verified working (15):** `MANIFEST`, `TEAM_TEASER`, `REASONS_VIDEO`, `VOTEBOX`, `VOTE_COUNTER`, `VOTE_RESULT`, `ELECTION`, `ELECTION_RESULT`, `ELECTION_RESULT_DIVERSITY`, `NEWSLETTER_SIGNUP`, `CLIMATE_LAB_COUNTER`, `CLIMATE_LAB_INLINE_TEASER`, `POSTCARD_GALLERY`, `CHALLENGE_ACCEPTED_INLINE_TEASER`, `COMPACT_DETAILS_FORM`.

Worth knowing: `VOTEBOX`, `ELECTION`, `NEWSLETTER_SIGNUP` and `CHALLENGE_ACCEPTED_INLINE_TEASER` all import `Account/Elements.js`, which has a `useRouter()` at line 126. They're safe only because they import `Hint`/`P` and never render the `AccountEnforceMe` export that contains it. That's luck, not design.

## How components were tested

**1. Static reachability analysis.** Walked the transitive import graph from all 24 entry points to find every `next/router`, `next/head` and `getInitialProps` occurrence, tracking the import chain to each. Necessary because *importing* `next/router` is harmless — only rendering the hook throws — so the grep alone gives false positives.

**2. Render verification under the real app router.** A temporary route rendered all 24 identifiers plus the 5 conditional sub-paths, each wrapped in an error boundary that recorded the thrown message, then asserted on the collected results in the browser.

- Before: **11 failures**, all `NextRouter was not mounted`
- After: **29 renders, 0 router errors**


apps/www/src/app/components/auth/login/code-form.tsx
```diff
@@ -39,8 +39,8 @@ export function CodeForm({
   redirectUrl,
 }: CodeFormProps) {
   const codeId = useId()
-  const router = useRouter()
-  const { query } = router
+  const searchParams = useSearchParams()
```
jstcki, line 42: Just a note of caution: using `useSearchParams` can cause a Next.js build issue, and this component _may_ need to be wrapped in `<Suspense>`

ovbm: Doesn't seem to apply here, since code-form is only ever rendered by apps/www/src/app/components/auth/login/index.tsx:75, which itself is only reachable from the App Router via the article/page Sanity renderer chain. That TRIAL_FORM identifier is loaded via next/dynamic(..., { ssr: false }) in DynamicComponents.tsx, so CodeForm/useSearchParams() never executes on the server at all — it only mounts client-side after hydration, which sidesteps the static-generation Suspense requirement entirely.

PR #1438, opened 2026-07-29
implement json-ld on article page

ovbm:
- moves social image resolving into it's own function
- new function that creates json-ld object and adds it to article pages


jstcki (approved):

jstcki:
looks good but the build should be fixed

